PR #30, opened 2016-02-24
feat(docs): add dockerfile

bacongobbler:


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.3.6
```
mboersma: Not `python:3.5.1-alpine`?

bacongobbler: that works; I'll change the image

PR #33, opened 2016-02-25
feat(managing_deis): add platform SSL documentation

### LINKED ISSUE #25: Document Platform SSL
https://github.com/deis/workflow/issues/322

bacongobbler:
closes #25 

pinging @krancour for review as he knows this field better than I do

### DIFF
--- a/src/managing-deis/platform-ssl.md
+++ b/src/managing-deis/platform-ssl.md
@@ -1,3 +1,62 @@
 # Platform SSL
 
-TODO (bacongobbler): rewrite for v2
+SSL/TLS is the standard security technology for establishing an encrypted link between a web server
+and a browser. This link ensures that all data passed between the web server and browsers remain
+private and integral.
+
+To enable SSL for your cluster and all apps running upon it, you can add an SSL key to your load
+balancer. You must either provide an SSL certificate that was registered with a CA or provide
+[your own self-signed SSL certificate](../reference-guide/creating_self_signed_ssl.md).
+
+## Installing SSL on a Load Balancer
+
+On most cloud-based load balancers, you can install a SSL certificate onto the load balancer
+itself. Any communication inbound to the load balancer will be encrypted while the internal
+components of Deis will still communicate over HTTP.
+
+When you install Deis, Kubernetes will provision a load balancer for the routers. To enable SSL,
+you will need to modify the listener settings on the load balancer:
+
+ - swap the load balancer protocol on port 443 to use HTTPS
+ - swap the backend protocol to use HTTP
+ - change the backend port to the same backend port as the listener on port 80
+ - install the certificate on the listener for port 443
+
+See your vendor's specific instructions on installing SSL on your load balancer. For AWS, see their
+documentation on [installing an SSL cert for load balancing](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/ssl-server-cert.html).
+
+## Installing SSL on the Deis Routers
+
+You can also use the Deis routers to terminate SSL connections. Use `kubectl` to install the
+certificate and private keys. Open your favorite text editor and create the Kubernetes manifest:
+
+	$ $EDITOR deis-router-default-cert.yaml
+
+At this point, you'll want to create a new Kubernetes secret bearing the wildcard certificate.
+The following criteria must be met:
+
+ - The name must be deis-router-default-cert
+ - The certificate's public key must be supplied as the value of the `cert` key
+ - The certificate's private key must be supplied as the value of the `key` key
+ - Both the certificate and private key must be base64 encoded
+
+For example:
+
+	$ cat deis-router-default-cert.yaml
+	apiVersion: v1
+	kind: Secret
+	metadata:
+	  name: deis-router-default-cert
+	  namespace: deis
+	type: Opaque
+	data:
+	  cert: LS0...tCg==
+	  key: LS0...LQo=
+
+If your certificate has intermediate certs that need to be presented as part of a certificate
+chain, append the intermediate certs to the bottom of the `cert` value before base64 encoding the
+cert chain.
+
+Once you've created the certificate manifest, you can then install the certificate with
+`kubectl create -f deis-router-default-cert.yaml`. The router will pick this up and update its
+configuration on-the-fly.


jchauncey:
This looks good to me. @helgi should look too

src/managing-deis/platform-ssl.md
```diff
@@ -1,3 +1,58 @@
...
+private and integral.
+
+To enable SSL for your cluster and all apps running upon it, you can add an SSL key to your load
+balancer. You must either provide an SSL certificate that was registered with a CA or provide
+[your own self-signed SSL certificate](../reference-guide/creating_self_signed_ssl.md).
+
+## Installing SSL on a Load Balancer
+
+On most cloud-based load balancers, you can install a SSL certificate onto the load balancer
+itself. Any communication inbound to the cluster will be encrypted while the internal components
```
krancour: That's slightly off.  If SSL terminates at the load balancer, the traffic entering the cluster is unencrypted.

src/managing-deis/platform-ssl.md
```diff
@@ -1,3 +1,58 @@
...
+of Deis will still communicate over HTTP.
+
+To enable SSL, you will need to open port 443 on the load balancer and forward it to port 80 on the
+routers. For AWS, you'll also need to add port 443 in the security group settings for your load
+balancer.
+
```
krancour: If doing it this way (using node ports) you'd need a health-check in place as well so that the load balancer only sends traffic to nodes that are actually running a router instance.  Otherwise, the high-numbered ports allocated dynamically by k8s should be used on the back end... unless these directions are strictly for _modifying_ the load balancer to support SSL and pre-suppose that the load balancer was set up correctly to begin with.

src/managing-deis/platform-ssl.md
```diff
@@ -1,3 +1,58 @@
...
+See your vendor's specific instructions on installing SSL on your load balancer. For AWS, see their
+documentation on [installing an SSL cert for load balancing](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/ssl-server-cert.html).
+
+## Installing SSL on the Deis Routers
```
krancour: This whole section is spot on.  :+1:

src/managing-deis/platform-ssl.md
bacongobbler: Would you prefer "Any communication inbound to the load balancer"?

src/managing-deis/platform-ssl.md
```diff
@@ -1,3 +1,58 @@
...
+
+You can also use the Deis routers to terminate SSL connections. Use `helm` to install the
+certificate and private keys. Open your favorite text editor with `helm edit`:
+
+	$ export EDITOR=vim
+	$ helm edit deis
+
+At this point, you'll want to create a new kubernetes secret bearing the wildcard certificate.
```
carmstrong: Should capitalize `Kubernetes`

src/managing-deis/platform-ssl.md
bacongobbler: These are directions to modify the load balancer to support SSL. I am assuming at this point after "installing Deis", we've already provisioned a kubernetes cluster that has a listener at port 80. This was all copied from the old documentation so I'll try to clarify that the user will need to add health checks as well for the node ports.

src/managing-deis/platform-ssl.md
krancour: Ya.  That would be accurate.

src/managing-deis/platform-ssl.md
krancour: > These are directions to modify the load balancer

I figured.

> I am assuming at this point after "installing Deis", we've already provisioned a kubernetes cluster that has a listener at port 80.

If running on AWS (which is where these docs are most applicable, because GKE does things differently) and k8s provisions the load balancer automatically, then there would also already be a listener on port 443-- although it would be connected to the high-numbered port on the kube-proxy that connects the the router's service's 443.  This would also be using TCP and not HTTPS.  (This is just how k8s would have set it up.)  So modifying the load balancer to terminate SSL is really a matter of:

1.  Swap the listener on 443 to use HTTPS instead of TCP.
2.  Swap the back end of that listener to use HTTP _and_ whatever back end port the listener on 80 is using.
3.  Install the cert.

krancour: And really, that's just _one_ way to do it.  The other way is to add healthchecks (which k8s would _not_ have done automatically) and then use the node port 80 on the back end.  This is actually slightly more efficient, network wise, because you avoid one hop through a kube-proxy by doing this since the _other_ configuration (with the high-numbered ports on the back end and no health check) could send the traffic to _any_ node and the node that receives the traffic might not actually be running a router.

krancour: Sorry... I'm going to shut up after this.  I think my point is that there's actually dizzying number of options for how to do this.  Part of me is tempted to drop this section entirely and have the docs deal _only_ with terminating SSL at the router since that requires less setup and is more secure.  If anyone really _wants_ to terminate at the load balancer, they can probably figure it out for themselves.

src/managing-deis/platform-ssl.md
```diff
@@ -1,3 +1,58 @@
...
+	data:
+	  cert: LS0...tCg==
+	  key: LS0...LQo=
+
+If your certificate has intermediate certs that need to be presented as part of a certificate
+chain, append the intermediate certs to the bottom of the sslCert value before base64 encoding the
+cert chain.
+
+Once you've created the certificate manifest, you can then re-install the platform with
+`helm uninstall deis && helm install deis`.
```
slack: Do we really want to recommend that folks reinstall the whole platform here?

bacongobbler: We could ask users to run `kubectl create -f deis-router-default-cert.yaml && kubectl --namespace=deis delete pod deis-router-abc123` but I felt that was too low-level for users. There's no equivalent command to reboot only the router until we can implement something like https://github.com/deis/charts/issues/1.

krancour: @bacongobbler, router actually does not require a reboot to see this change.  :)  Just adding the cert secret is enough.

bacongobbler: Oh excellent! I'll remove this part then.

src/managing-deis/platform-ssl.md
bacongobbler: > So modifying the load balancer to terminate SSL is really a matter of:
> 
>     Swap the listener on 443 to use HTTPS instead of TCP.
>     Swap the back end of that listener to use HTTP and whatever back end node port the listener on 80 is already using.
>     Install the cert.

Perfect! That's all I was trying to convey. I'd like to use the host port, however I feel like we cannot rely on this behaviour. For example, what if a user's application (or another third party app) binds to port 80 on the host? Given that there's enough nodes to schedule that pod, a load balancer pointing at the minion's host port 80 would then cycle this user's app into the routing table because it is listening on port 80, passing the health check. I think it's best to play around with the load balancer that kubernetes has handed us :)

> I think my point is that there's actually a dizzying number of options for how to do this.

That is certainly true. There are many networking topologies; I guess it depends on how far down the network stack we want to stop (kube-proxy vs host port). I tend to lean towards the former, even if there's a performance hit for now.

> Part of me is tempted to drop this section entirely and have the docs deal only with terminating SSL at the router since that requires less setup and is more secure. If anyone really wants to terminate at the load balancer, they can probably figure it out for themselves.

We said that initially, then some users care whether SSL is terminated at the load balancer vs. termination at the routers. I think it's best to at least point users in the right direction. :)

bacongobbler:
All changes addressed and ready for another round of reviews.

src/managing-deis/platform-ssl.md
krancour: > I'd like to use the host port, however I feel like we cannot rely on this behaviour. For example, what if a user's application (or another third party app) binds to port 80 on the host? Given that there's enough nodes to schedule that pod, a load balancer pointing at the minion's host port 80 would then cycle this user's app into the routing table because it is listening on port 80, passing the health check.

That is an excellent point, but there is also a confluence of three specific things required to make that happen-- 1. the app would have to bind to host port 80, 2. it would have to bind to host port 9090 3. it would have to use `:9090/healthz` for the health check.  I figured this combination of things was extraordinarily unlikely primarily on account of the fact that user applications are almost guaranteed to listen for health checks on the same port on which they listen for everything else.  The only reason the router splits health checks out into a separate port is because I had to provide a listener on the Nginx end that was guaranteed to _never_ use PROXY protocol, since ELBs, for instance, could only do TCP-based health checks with PROXY proto in play and I wanted HTTP based health checks so that we could call out the /healthz endpoint specifically.

In the end, you are correct that this _could_ happen.  Perhaps it's fair that we not _recommend_ using the router's host ports, but we ourselves should remain tuned in to the fact that it's a viable option with slightly better performance.

Ultimately, I think if your directions are just to amend the LB's 443 listener to use HTTPS on the front and the HTTPS on the back with the same back-end port as the 80 listener (which should be a node port), that will be fine.

src/managing-deis/platform-ssl.md
```diff
@@ -1,3 +1,62 @@
...
+[your own self-signed SSL certificate](../reference-guide/creating_self_signed_ssl.md).
+
+## Installing SSL on a Load Balancer
+
+On most cloud-based load balancers, you can install a SSL certificate onto the load balancer
+itself. Any communication inbound to the load balancer will be encrypted while the internal
+components of Deis will still communicate over HTTP.
+
+When you install Deis, Kubernetes will provision a load balancer for the routers. To enable SSL,
+you will need to modify the listener settings on the load balancer:
```
krancour: This is only true on certain underlying infrastructure.  e.g. true on AWS and GKE, but valse on bare metal or vagrant.  It is possibly a good idea to mention that and to also mention that if running on a platform where the load balancer was not automatically provisioned by k8s, generic, manual load balancer configuration instructions are available here: https://github.com/deis/router#manually-configuring-a-load-balancer

Your doc here could instruct readers to resume these instructions after they have set up their load balancer.

krancour: Although, the previous paragraph says "most cloud providers," so maybe you already covered your bases and narrowed the scope of the discussion.

Up to you if you want to make any further clarification here or not.  It need not stand in the way of my LGTM.

src/managing-deis/platform-ssl.md
```diff
@@ -1,3 +1,62 @@
...
+	  cert: LS0...tCg==
+	  key: LS0...LQo=
+
+If your certificate has intermediate certs that need to be presented as part of a certificate
+chain, append the intermediate certs to the bottom of the `cert` value before base64 encoding the
+cert chain.
+
+Once you've created the certificate manifest, you can then install the certificate with
+`kubectl create -f deis-router-default-cert.yaml`. The router will pick this up and update its
+configuration on-the-fly.
```
krancour: :+1:

src/managing-deis/platform-ssl.md
bacongobbler: let's follow up with that after we merge this :)

PR #36, opened 2016-02-26
ref(APIv1): remove v1 controller docs

slack:
Probably don't need this in the v2 repo.


bacongobbler:
You're probably right. I was thinking that this will eventually become docs.deis.io and v1 users won't have any documentation, but then again they can just check out an older version of Deis. There's no need to keep a significantly older version's documentation available here.

PR #41, opened 2016-02-29
doc(install): fixup helm invocations

### LINKED ISSUE #15: Install instructions require update
I believe we need to call out a newer version of helm.  (Docs currently show 0.2) and that we also need to include the `helm generate` step in the install process.

slack:
Update install instructions to include `helm generate` step.

Fixes #15

### DIFF
--- a/src/installing-deis/installing-deis-workflow.md
+++ b/src/installing-deis/installing-deis-workflow.md
@@ -47,7 +47,7 @@ Now that you have Helm installed and added the Deis Chart Repository, install Wo
 
 ```
 $ helm generate -x manifests deis      # generates various secrets
-$ helm install deis --namespace=deis   # injects resources into your cluster
+$ helm install deis                    # injects resources into your cluster
 ```
 
 Helm will install a variety of Kubernetes resources in the `deis` namespace.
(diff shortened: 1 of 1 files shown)


src/installing-deis/installing-deis-workflow.md
```diff
@@ -46,7 +46,8 @@ $ helm fetch deis/deis
 Now that you have Helm installed and added the Deis Chart Repository, install Workflow by running:
 
 ```
-$ helm install deis/deis --namespace=deis
+$ helm generate deis                   # generates various secrets
```
mboersma: Maybe this should be `helm generate -x manifests deis`. That's the current way to work around the double-templating.

slack: Ah, ok. Will update.

src/installing-deis/installing-deis-workflow.md
```diff
@@ -46,7 +46,8 @@ $ helm fetch deis/deis
 Now that you have Helm installed and added the Deis Chart Repository, install Workflow by running:
 
 ```
-$ helm install deis/deis --namespace=deis
+$ helm generate -x manifests deis      # generates various secrets
+$ helm install deis --namespace=deis   # injects resources into your cluster
```
mboersma: We don't need to specify `--namespace=deis` here since the chart includes it. (It's only `helm uninstall` that needs hand-holding.)

slack: Ah, ok.

krancour:
@slack isn' there a missing `helm fetch deis/deis` here?

PR #43, opened 2016-02-29
doc(diagrams): remove old diagrams, use new git workflow

slack:


mboersma:
Changes look good although I wasn't able to test them directly (see #45 for the fix to that).

PR #46, opened 2016-02-29
doc(installing): Minor refactor of install docs

krancour:
This improves the clarity of the installation instructions a little bit by moving the instructions re: the `helm fetch` command from the section that documents adding the deis/charts repo to the section that documents the other install-related helm commands.

I also removed the `helm update` command which seems like an irrelevant operation wrt to installing a package from a chart repo we've _just_ added.


krancour:
CI failure is not related to these changes.  Merging.

PR #58, opened 2016-03-01
bug(typo): s/deatch/detach/

slack:


mboersma:
See also deis/workflow#446.

PR #59, opened 2016-03-02
doc(installing): Don't always monospace helm

krancour:
Another nit...

Reading through the docs, I felt assaulted by the continuous use of `monospacing` in reference to Helm.  It looks awkward in places where it's referring to the _product_ and not referring explicitly to either a _command_ that someone is meant to or to output _from_ some command.

By fixing this and also upcasing it to "Helm," Helm gets a little more respect as a product-- which it deserves.

I also changed the link that read "installing helm" to "Helm's own documentation," because I felt that was more appropriate, given that we're not linking to another section of the Workflow documentation, but rather to something external.


jchauncey:
@krancour this is ready to merge

PR #63, opened 2016-03-02
doc(process types): Eliminate repition and generally polish

krancour:
There was a lot of repetition of information in the process type docs.  I think it told me _three_ times that the `Procfile` goes in the root of my application's source tree.  I've edited this page to be a bit more concise.

### DIFF
--- a/src/using-deis/process-types-and-the-procfile.md
+++ b/src/using-deis/process-types-and-the-procfile.md
@@ -1,23 +1,25 @@
 # Process Types and the Procfile
 
-A Procfile is a mechanism for declaring what commands are run by your application’s containers on
-the Deis platform. It follows the [process model][]. You can use a Procfile to declare various
-process types, such as multiple types of workers, a singleton process like a clock, or a consumer
-of the Twitter streaming API.
+Deis Workflow observes the [process model][].  As part of this model, a Procfile may be used as a
+mechanism for declaring the commands that should be run by your application’s containers.  You can
+use a Procfile to declare multiple processes for different types of application workers-- for
+example, an application server, a job execution timer, or a consumer of the Twitter streaming API.
 
-## Process Types as Templates
+Multiple process types can be scaled independently of one another.
 
-A Procfile is a text file named `Procfile` placed in the root of your application that lists the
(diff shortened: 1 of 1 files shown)


slack:
Great changes. If you have a `--procfile` cli invocation handy it would be great to add that to the documentation, the syntax is not obvious.

krancour:
@slack added an example of `deis pull` with `--procfile` option, as you suggested.

slack:
Oh snap, I thought you had to pass in a YAML formatted string to `--procfile="---\n..."`, not a file ref.

krancour:
@slack I didn't try it... now you have me second-guessing.  I will be sure to try it before I merge this.

krancour:
@slack, you were right.  I'll have to fix it.

```
$ deis help pull
Creates a new build of an application. Imports an <image> and deploys it to Deis
as a new release. If a Procfile is present in the current directory, it will be used
as the default process types for this application.

Usage: deis builds:create <image> [options]

Arguments:
  <image>
    A fully-qualified docker image, either from Docker Hub (e.g. deis/example-go:latest)
    or from an in-house registry (e.g. myregistry.example.com:5000/example-go:latest).
    This image must include the tag.

Options:
  -a --app=<app>
    The uniquely identifiable name for the application.
  -p --procfile=<procfile>
    A YAML string used to supply a Procfile to the application.
```

krancour:
@slack testing this out seems to have exposed a bug... seems process types other than the defaults (web for buildpacks and cmd otherwise) are not working at all.  It makes it hard to verify this functionality to include in the docs.  Going to open an issue in workflow and hold off on merging this for now.

krancour:
Actually, it seems this issue is already captured by https://github.com/deis/workflow/issues/420.

krancour:
I'm removing LGTMs to avoid this PR appearing at a glance to be mergeable.

bacongobbler:
I think adding docs here are fine as long as that's the intentional behaviour. We'll fix up deis/workflow#420 to match the documentation.

krancour:
Per @bacongobbler's suggestion, this can be re-reviewed and merged, I think.

PR #64, opened 2016-03-03
doc(configuring-dns): Tip about not forgetting trailing slash

ultimateboy:
Sometimes the little things (like a trailing slash) take longer to figure out than you want to admit. Hopefully this saves somebody else some troubles :)


slack:
You and me both! Thanks for the fix!

PR #65, opened 2016-03-04
doc(buildpacks): Fix broken builder link

krancour:


src/using-deis/using-buildpacks.md
```diff
@@ -122,6 +122,7 @@ and set `BUILDPACK_URL` to the URL:
 
 [pods]: http://kubernetes.io/v1.1/docs/user-guide/pods.html
 [controller]: ../understanding-deis/components.md#controller
+[builder]: ../understanding-deis/components.md#controller
```
bacongobbler: should this point to a `#builder` internal link instead? :)

krancour: Whoops.  Copy/paste error.  Thanks for catching.

krancour:
@bacongobbler, fixed.  Thanks again for the catch.

PR #66, opened 2016-03-04
doc(using dockerfiles): Apply minor cleanup

krancour:
Nits... This PR monospaces Dockerfile directives, which seems more proper, and just generally improves the precision of language with respect to these requirements.


jchauncey:
@krancour this is ready to merge

PR #68, opened 2016-03-05
doc(configuring-object-storage.md): add section on storing credentials & GCS

arschles:
also adds the database component to the list of things that needs S3

Fixes deis/builder#197

cc/ @smothiki

### DIFF
--- a/src/installing-deis/configuring-object-storage.md
+++ b/src/installing-deis/configuring-object-storage.md
@@ -23,6 +23,10 @@ The Deis components determine what object storage system to use via environment 
   - `storage.googleapis.com` for Google Cloud Storage
 - `DEIS_MINIO_SERVICE_HOST` and `DEIS_MINIO_SERVICE_PORT` - The in-cluster Minio service. Note that these will be set automatically by Kubernetes if you run [Minio](http://minio.io) as a service in the cluster. See [the Minio service from the Deis Minio Chart](https://github.com/deis/charts/blob/master/deis-dev/manifests/deis-minio-service.yaml) for an example service.
 
+## Specifying the Bucket
+
+[deis/builder](https://github.com/deis/builder) uses an additional environment variable, `BUCKET` to determine the name of the bucket (in the specified object storage system) to use. It uses `git` as the default bucket name, but if your credentials (see below) don't have read and write access to it, you'll have to specify a different bucket. To do so, simply set the `BUCKET` environment variable to another value (`deis-builds`, for example).
+
 # Storing Credentials
 
 In the Deis V2 Beta release, all components read credentials from the filesystem, and we suggest that credentials are stored in [Kubernetes secrets](http://kubernetes.io/v1.1/docs/user-guide/secrets.html) and mounted to the appropriate location for the component. See the below list for the expected location for each component, and see [the deis-dev chart](https://github.com/deis/charts/tree/master/deis-dev) for examples of using and mounting secrets.
--- a/src/installing-deis/configuring-object-storage.md
+++ b/src/installing-deis/configuring-object-storage.md
@@ -98,9 +98,19 @@ The registry reads credentials from the below locations on the filesystem.
 
 ## [deis/database](https://github.com/deis/postgres)
 
+The database is configured slightly differently from the other components. It looks for an environment variable which it then uses as a key to look up the object storage configuration in a config file.
+
 ### Environment Variables
 
-TODO
+The database looks for the `DATABASE_STORAGE` environment variable to determine what object storage system to use. Valid values are listed below and the [Helm chart for Deis](https://github.com/deis/charts/tree/master/deis-dev) defaults to `minio`:
+
+- filesystem: Store persistent data on ephemeral disk
+- s3: Store persistent data in AWS S3 (configure in S3 section)
+- azure: Store persistent data in Azure's object storage
+- gcs: Store persistent data in Google Cloud Storage
+- minio: Store persistent data on in-cluster Minio server
+
+(TODO: is this correct?)
 
 ### Credentials
 
@@ -108,6 +118,8 @@ The database reads credentials from the below locations on the filesystem.
 
 - Key: `/etc/wal-e.d/env/access-key-id`
 - Secret: `/etc/wal-e.d/env/access-key-secret`
+
+(TODO: is this still correct?)
 
 
 # Limitations
--- a/src/installing-deis/configuring-object-storage.md
+++ b/src/installing-deis/configuring-object-storage.md
@@ -8,51 +8,110 @@ A variety of Deis components rely on an object storage system to do their work. 
 - [registry](https://github.com/deis/registry)
 - [database](https://github.com/deis/postgres)
 
-These components are built flexibly, so they can work out of the box with almost any system that is compatible with the [S3 API](http://docs.aws.amazon.com/AmazonS3/latest/API/APIRest.html).
+These components are flexible and can work out of the box with almost any system that is compatible with the [S3 API](http://docs.aws.amazon.com/AmazonS3/latest/API/APIRest.html).
 
 # Minio
 
 Additionally, Deis ships with a [Minio](http://minio.io) [component](https://github.com/deis/minio). This component runs as a Kubernetes service, and the components listed above are configured to automatically look for that service and use it as object storage if it's available.
 
-# Telling Deis What to Use
+# Configuring the Deis Components
 
-The Deis components determine what object storage system to use via environment variables that you set up. The below list is the lookup order for all Deis components.
+Every Deis component that relies on object storage relies on the following two inputs for configuration:
+
+- One or more environment variables with host and port to describe where the object storage system is
+- One or more files to provide access credentials for the object storage system.
+	- We suggest storing these values in [Kubernetes secrets](http://kubernetes.io/v1.1/docs/user-guide/secrets.html) and mounting them as volumes to each pod
+	- See [the deis-dev chart](https://github.com/deis/charts/tree/master/deis-dev) for examples of using and mounting secrets.
+
+The subsections herein explain how to configure these two inputs for each applicable component.
+
+## [deis/builder](https://github.com/deis/builder)
+
+### Environment Variables
+
+The builder looks for the below environment variables to determine where the object storage system is. The builder looks in-order for these variables. If it finds two, the one higher in the list will be used.
 
 - `DEIS_OUTSIDE_STORAGE` - The external S3-compatible object storage system. Commonly used URLs:
   - `s3.amazonaws.com` for Amazon S3's `us-east-1a` region
   - `storage.googleapis.com` for Google Cloud Storage
-- `DEIS_MINIO_SERVICE_HOST` and `DEIS_MINIO_SERVICE_PORT` - The in-cluster Minio service. Note that these will be set automatically by Kubernetes if you run [Minio](http://minio.io) as a service in the cluster. See [the Minio service from the Deis Minio Chart](https://github.com/deis/charts/blob/master/deis-dev/manifests/deis-minio-service.yaml) for an example service.
+- `DEIS_MINIO_SERVICE_HOST` and `DEIS_MINIO_SERVICE_PORT` - The in-cluster Minio service. Additional notes about these variables:
+  - They are set automatically by Kubernetes if you run [Minio](http://minio.io) as a service in the cluster
+  - The [Helm chart for Deis](https://github.com/deis/charts/tree/master/deis-dev) installs Minio by default, so the Builder will use Minio by default.
 
-## Specifying the Bucket
+The builder also uses an environment variable to determine the name of the bucket it should store build artifacts in. It uses `git` by default, but if your credentials (see below) don't have read and write access to it, you'll have to specify a different bucket. To do so, simply set the `BUCKET` environment variable to another value (`deis-builds`, for example).
 
-[deis/builder](https://github.com/deis/builder) uses an additional environment variable, `BUCKET` to determine the name of the bucket (in the specified object storage system) to use. It uses `git` as the default bucket name, but if your credentials (see below) don't have read and write access to it, you'll have to specify a different bucket. To do so, simply set the `BUCKET` environment variable to another value (`deis-builds`, for example).
+### Credentials
 
-# Storing Credentials
+The builder reads credentials from the below locations on the filesystem.
 
-In the Deis V2 Beta release, all components read credentials from the filesystem, and we suggest that credentials are stored in [Kubernetes secrets](http://kubernetes.io/v1.1/docs/user-guide/secrets.html) and mounted to the appropriate location for the component. See the below list for the expected location for each component, and see [the deis-dev chart](https://github.com/deis/charts/tree/master/deis-dev) for examples of using and mounting secrets.
+- Key: `/var/run/secrets/object/store/access-key-id`
+- Secret `/var/run/secrets/object/store/access-key-secret`
 
-- [builder](https://github.com/deis/builder)
-  - Key: `/var/run/secrets/object/store/access-key-id`
-  - Secret `/var/run/secrets/object/store/access-key-secret`
-- [slugbuilder](https://github.com/deis/slugbuilder)
-  - Key: `/var/run/secrets/object/store/access-key-id`
-  - Secret `/var/run/secrets/object/store/access-key-secret`
-- [slugrunner](https://github.com/deis/slugrunner)
-  - Key: `/var/run/secrets/object/store/access-key-id`
-  - Secret: `/var/run/secrets/object/store/access-key-secret`
-- [registry](https://github.com/deis/registry)
-  - Key: `/var/run/secrets/deis/registry/creds/accesskey`
-  - Secret: `/var/run/secrets/deis/registry/creds/secretkey`
-- [database](https://github.com/deis/postgres)
-  - Key: `/etc/wal-e.d/env/access-key-id`
-  - Secret: `/etc/wal-e.d/env/access-key-secret`
+### A Note on Google Cloud Storage
 
-# A Note on Google Cloud Storage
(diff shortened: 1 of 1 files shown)


src/installing-deis/configuring-object-storage.md
```diff
@@ -20,6 +21,32 @@ The Deis components determine what object storage system to use via environment
...
+  - Secret: `/var/run/secrets/deis/registry/creds/secretkey`
+- [database](https://github.com/deis/postgres)
+  - Key: `/etc/wal-e.d/env/access-key-id`
+  - Secret: `/etc/wal-e.d/env/access-key-secret`
+
+# A Note on Google Cloud Storage
+
+As you may know Google Cloud Storage (GCS) can [interoperate with the S3 API](https://cloud.google.com/storage/docs/interoperability), and, if you choose to use Google Cloud Storage for object storage, you'll have to turn on this interoperability mode.
+
+If you choose to use Google Cloud Storage, set your `DEIS_OUTSIDE_STORAGE_HOST` environment variable to `storage.googleapis.com`, and follow [these instructions](https://cloud.google.com/storage/docs/migrating?hl=en_US#keys) to generate an S3 compatible access key ID and access key secret. Store these credentials just as you would if they were AWS S3 or Minio credentials (see the "Storing Credentials" section above).
```
smothiki: URL should be along with the scheme http or https 
https://github.com/deis/builder/blob/master/pkg/gitreceive/storage/endpoint.go#L25

smothiki:
Builder needs a separate config settings for any endpoint same instructions goes for aws S3 as well.
Also BUCKET is the env variable that needs to be set, bucket names are Global and defaults to "git" which will fail

smothiki:
Please don't merge this until documentation is complete.

bacongobbler:
> Also BUCKET is the env variable that needs to be set, bucket names are Global and defaults to "git" which will fail

I think that can be specified elsewhere in another PR since this just addresses S3 credentials.

smothiki:
Builder needs a separate configuration apart from other components in the platform. Also there is an error in the docs, need to set URL scheme as well

src/installing-deis/configuring-object-storage.md
arschles: @smothiki if the URL doesn't have a scheme, then that `stripScheme` `func` is a no-op

arschles:
@smothiki:

>Builder needs a separate config settings for any endpoint same instructions goes for aws S3 as well.

I don't understand what you mean here. What is the separate config setting besides `DEIS_OUTSIDE_STORAGE_HOST`?

>Also BUCKET is the env variable that needs to be set, bucket names are Global and defaults to "git" which will fail

Why will `git` fail?

>Builder needs a separate configuration apart from other components in the platform.

Similar to my above comment, can you give more details on what the separate configuration is? IIRC, `DEIS_OUTSIDE_STORAGE_HOST` and the credentials setup mentioned in these docs is enough.

I've removed the `LGTM2` label from this PR until these questions are sorted out.

src/installing-deis/configuring-object-storage.md
smothiki: ok also https://github.com/deis/builder/blob/master/pkg/gitreceive/storage/endpoint.go#L13
its  "DEIS_OUTSIDE_STORAGE" there is no HOST

smothiki:
if there is a git bucket already exists in GCS storage we cannot PUT or GET from the bucket if "git " bucket is not registered under user credentials

smothiki:
Also the code defaults to MInio USER host and PORT . So if user has to set back-end to AWS . He has to do the similar config changes as GCS but it is not documented in this DOC

src/installing-deis/configuring-object-storage.md
arschles: Fixed in 1a0ba7024497851c8594c8edd09eaa0a134f2f34

arschles:
@smothiki I've added docs for the `BUCKET` env var in 6cbee9bebe8dc7426b978e62c32646d8c9579595

arschles:
@smothiki according to https://github.com/deis/builder/blob/master/pkg/gitreceive/storage/endpoint.go#L53-L62, the builder defaults to outside storage. Do my recent changes in https://github.com/arschles/deis-docs-v2/commit/1a0ba7024497851c8594c8edd09eaa0a134f2f34 clearly specify that the `DEIS_OUTSIDE_STORAGE` env var applies to either S3 or GCS?

smothiki:
I meant deis-builder-rc in deis-dev charts  by default doesn't have DEIS_OUTSIDE_STORAGE variable set. So code defaults to Minio host and port. 
Also  s3.amazonaws.com works only for us-east-1 . If user wants to specify for some other region he has to prefix the uRL with s3-<region-name> .

smothiki:
I think you have generalized this for every deis component. The setting are only valid for builder and its components not for other services. ping @bacongobbler for database , @kmala for registry.

src/installing-deis/configuring-object-storage.md
```diff
@@ -15,10 +16,42 @@ Additionally, Deis ships with a [Minio](http://minio.io) [component](https://git
 
 # Telling Deis What to Use
 
-The Deis components determine what object storage system to use via environment variables that you set up:
+The Deis components determine what object storage system to use via environment variables that you set up. The below list is the lookup order for all Deis components.
```
smothiki: @arschles Im good with the PR if you can some how add a special column for builder storage settings. This heading says the below settings are for deis which is not true.

arschles:
@smothiki I've separated each component into its own section in https://github.com/arschles/deis-docs-v2/commit/238a71f7ff112a18f77ee911e42f9c8b9ce6f3db. cc/ @kmala and @bacongobbler

src/installing-deis/configuring-object-storage.md
arschles: yup, added sections in https://github.com/arschles/deis-docs-v2/commit/238a71f7ff112a18f77ee911e42f9c8b9ce6f3db

arschles:
@smothiki also, I've clarified that `s3.amazonaws.com` is specifically only for `us-east-1a` in https://github.com/arschles/deis-docs-v2/commit/3b1eb4a84bb70ef81f3ff3bfa9cf6dd49eee3361

arschles:
and finally, @smothiki I've noted at https://github.com/arschles/deis-docs-v2/commit/238a71f7ff112a18f77ee911e42f9c8b9ce6f3db#diff-626fcb4e0383b6bc5899531540a77d1aR39 that the `deis-dev` chart defaults to using Minio

smothiki:
DEIS registry and Database follows same configuration. So we should deal with that in a different PR. I'll open an issue for that . Giving this an LGTM

arschles:
@smothiki ok, thanks. For now, I'd like to leave everything in its own section. If/when we move to a unified object storage configuration convention (I think we should), then I'm happy to merge them all together.

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,124 @@ A variety of Deis components rely on an object storage system to do their work.
...
+
+The database looks for the `DATABASE_STORAGE` environment variable to determine what object storage system to use. Valid values are listed below and the [Helm chart for Deis](https://github.com/deis/charts/tree/master/deis-dev) defaults to `minio`:
+
+- filesystem: Store persistent data on ephemeral disk
+- s3: Store persistent data in AWS S3 (configure in S3 section)
+- azure: Store persistent data in Azure's object storage
+- gcs: Store persistent data in Google Cloud Storage
+- minio: Store persistent data on in-cluster Minio server
+
+(TODO: is this correct?)
```
arschles: @bacongobbler can you comment on this and the below `TODO`?

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,124 @@ A variety of Deis components rely on an object storage system to do their work.
...
+The slugrunner reads credentials from the below locations on the filesystem.
+
+- Key: `/var/run/secrets/object/store/access-key-id`
+- Secret: `/var/run/secrets/object/store/access-key-secret`
+
+## [deis/registry](https://github.com/deis/registry)
+
+### Environment Variables
+
+TODO
```
arschles: @kmala can you comment on this TODO and whether the below `Credentials` section is correct?

kmala:
@arschles i had commented on this link https://gist.github.com/arschles/5b35a742028f2c784f00#registry
Can you update this doc with that

src/installing-deis/configuring-object-storage.md
bacongobbler: The only two supported at this time are `s3` and "anything other than s3", which is minio.

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,124 @@ A variety of Deis components rely on an object storage system to do their work.
...
+(TODO: is this correct?)
+
+### Credentials
+
+The database reads credentials from the below locations on the filesystem.
+
+- Key: `/etc/wal-e.d/env/access-key-id`
+- Secret: `/etc/wal-e.d/env/access-key-secret`
+
+(TODO: is this still correct?)
```
bacongobbler: yes, this is still correct

arschles:
@kmala made the changes you requested in https://github.com/arschles/deis-docs-v2/commit/b297bc79cd4d82a297bb7e995f323c6e2a949cea

src/installing-deis/configuring-object-storage.md
arschles: @bacongobbler I made a few updates from what I inferred from how the `deis-objectstorage` secret works and how the [database RC manifest](https://github.com/deis/charts/blob/master/deis-dev/tpl/deis-database-rc.yaml) is laid out. Mind checking things out once more?

arschles:
@kmala @bacongobbler @smothiki I've removed both LGTMs, since after they were both on, there were a significant amount of comments changes. Can you all do another review?

bacongobbler:
database documentation is a wee bit generic, but it's great that we have this in here so we can point users where to look. :+1:

arschles:
@bacongobbler thanks for the honesty (I suspected as much!) - think this is ok to merge, along with an issue to improve these docs?

bacongobbler:
absolutely. This is a great starting point that we can build on top of.

arschles:
k, cool. thanks!

bacongobbler:
closes deis/postgres#29

bacongobbler:
also closes https://github.com/deis/docs-v2/issues/26

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,109 @@ A variety of Deis components rely on an object storage system to do their work.
...
+## Credentials
+
+The database reads the credentials information from a `/var/run/secrets/deis/objectstore/creds/objectstorage-keyfile` file. See https://github.com/deis/charts/blob/master/deis-dev/tpl/deis-objectstorage-secret.yaml for an example of what that file should look like.
 
 # Limitations
 
-The only currently known limitation is that [the Deis registry component](https://github.com/deis/registry) will not automatically look up the minio service, nor will it look for other storage env vars. That fix is being tracked in a [GitHub issue](https://github.com/deis/registry/issues/7) and is planned for our beta release.
+Below is a list of known limitations of our components' ability to interact with object storage systems.
+
+- [The Deis registry component](https://github.com/deis/registry) will not automatically look up the Kubernetes Minio service, nor will it look for other storage env vars. That fix is being tracked in a [GitHub issue](https://github.com/deis/registry/issues/7) and is planned for our beta release.
```
kmala: This is no more a limitation.fixed in https://github.com/deis/registry/pull/27

arschles: thanks @kmala. removed in https://github.com/arschles/deis-docs-v2/commit/7851b48f5b6c6cca45a9cc1093390a2f23c9e352

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,103 @@ A variety of Deis components rely on an object storage system to do their work.
...
+
+The registry is configured slightly differently from most of the other components. Read on for details.
+
+### Environment Variables
+
+The registry looks for a `REGISTRY_STORAGE` environment variable, which it then uses as a key to look up the object storage location and authentication information in a configuration file. See below for details on that file.
+
+### Credentials
+
+The registry reads the credential information from a `/var/run/secrets/deis/registry/creds/objectstorage-keyfile` file. See https://github.com/deis/charts/blob/master/deis-dev/tpl/deis-objectstorage-secret.yaml for an example of what that file should look like.
```
kmala: Can we write this way: The registry reads the credential information from a `/var/run/secrets/deis/registry/creds/objectstorage-keyfile` file. This is generated automatically during `helm generate` based on the configuration options given in the https://github.com/deis/charts/blob/master/deis-dev/tpl/objectstorage.toml.

arschles: @kmala ok, changed it in d106a7eb245692aac7d5f12b1e2e9a77b34a8653. Also changed the DB credentials section to have similar language in fb6131dbaedfb985b3c7eb0c06ef4be57c688e46

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,103 @@ A variety of Deis components rely on an object storage system to do their work.
...
 
-# Limitations
+The subsections herein explain how to configure these two inputs for each applicable component.
 
-The only currently known limitation is that [the Deis registry component](https://github.com/deis/registry) will not automatically look up the minio service, nor will it look for other storage env vars. That fix is being tracked in a [GitHub issue](https://github.com/deis/registry/issues/7) and is planned for our beta release.
+## [deis/builder](https://github.com/deis/builder)
+
+### Environment Variables
+
+The builder looks for the below environment variables to determine where the object storage system is. The builder looks in-order for these variables. If it finds two, the one higher in the list will be used.
```
jackfrancis: It might be easier to just explicitly state the preferred variable in the case of all of them being present. E.g.,

If it finds two, `DEIS_OUTSIDE_STORAGE` will take precedence and external object storage configuration will be applied.

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,103 @@ A variety of Deis components rely on an object storage system to do their work.
...
+### Credentials
+
+The builder reads credentials from the below locations on the filesystem.
+
+- Key: `/var/run/secrets/object/store/access-key-id`
+- Secret `/var/run/secrets/object/store/access-key-secret`
+
+### A Note on Google Cloud Storage
+
+As you may know, Google Cloud Storage (GCS) can [interoperate with the S3 API](https://cloud.google.com/storage/docs/interoperability), and, if you choose to use Google Cloud Storage for object storage, you'll have to turn on this interoperability mode.
```
jackfrancis: "As you may know" isn't really necessary and may be off-putting to any readers who don't in fact know this.

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,103 @@ A variety of Deis components rely on an object storage system to do their work.
...
+If you choose to use Google Cloud Storage, set your `DEIS_OUTSIDE_STORAGE` environment variable to `storage.googleapis.com`, and follow [these instructions](https://cloud.google.com/storage/docs/migrating?hl=en_US#keys) to generate an S3 compatible access key ID and access key secret. Store these credentials just as you would if they were AWS S3 or Minio credentials. As mentioned above, we recommend storing these as Kubernetes secrets. See the "Configuring Deis Components" section above for more details and examples.
+
+## [deis/slugbuilder](https://github.com/deis/slugbuilder)
+
+### Environment Variables
+
+The slugbuilder looks for the below environment variables to determine where to download code from and upload slugs to.
+
+- `TAR_URL` - The location of the `.tar` archive (which it will build)
+- `put_url` - The location this component will upload the finished slug to
```
jackfrancis: `put_url` should be expressed as `PUT_URL` for consistency (unless there's an idiom for this particular variable that I'm overlooking).

src/installing-deis/configuring-object-storage.md
```diff
@@ -6,20 +6,103 @@ A variety of Deis components rely on an object storage system to do their work.
...
+
+The registry is configured slightly differently from most of the other components. Read on for details.
+
+### Environment Variables
+
+The registry looks for a `REGISTRY_STORAGE` environment variable, which it then uses as a key to look up the object storage location and authentication information in a configuration file. See below for details on that file.
+
+### Credentials
+
+The registry reads the credential information from a `/var/run/secrets/deis/registry/creds/objectstorage-keyfile` file. This is generated automatically during helm generate based on the configuration options given in the https://github.com/deis/charts/blob/master/deis-dev/tpl/objectstorage.toml.
```
jackfrancis: surround "helm generate" in backticks so the sentence looks like:

This is generated automatically during `helm generate` based on the configuration...

src/installing-deis/configuring-object-storage.md
arschles: I unfortunately can't. The slugbuilder is implemented to look for `put_url` right now. We should change that in the future, but that's out of scope of this PR.

src/installing-deis/configuring-object-storage.md
arschles: Good point, fixed in https://github.com/deis/docs-v2/commit/a5aa1329018fa54398673975bca3426d36888cf4

src/installing-deis/configuring-object-storage.md
arschles: Fixed in 30527a1cf2fce275137dae9638a1168cd8989611

src/installing-deis/configuring-object-storage.md
arschles: Good idea. Fixed in https://github.com/deis/docs-v2/commit/f575341945ada4b0ce5fd59c52622d9e56a50f2f

jackfrancis:
All amendments look good!

arschles:
thanks @kmala @smothiki @jackfrancis @bacongobbler for your reviews

PR #71, opened 2016-03-06
core(docs): callout doc version explicitly

slack:
Build a callout so folks aren't confused between the different versions of
documentation.

image

### DIFF
--- a/themes/deis/base.html
+++ b/themes/deis/base.html
@@ -59,6 +59,11 @@
             </div>
             <div class="span6 column_calc contents">
                 <div class="doc-content">
+                    <div class="version-warning important">
+                        <p class="version-warning-title">Version 2</p>
+                        <p>This documentation is for Deis Workflow (v2). For v1 documentation visit <a href="http://docs.deis.io/en/latest/">http://docs.deis.io/en/latest/</a>.</p>
+                    </div>
+
                     {{ content }}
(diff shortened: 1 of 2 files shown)


slack:
If we are okay with this approach, I may PR against the v1 docs adding a corresponding notice once v2 docs have a semi-stable home.

themes/deis/base.html
```diff
@@ -59,6 +59,11 @@
             </div>
             <div class="span6 column_calc contents">
                 <div class="doc-content">
+                    <div class="version-warning important">
+                        <p class="version-warning-title">Version 2</p>
+                        <p>This documentation is for Deis Workflow (v2), for v1 documentation visit <a href="http://docs.deis.io/en/latest/">http://docs.deis.io/en/latest/</a>.</p>
```
mboersma: I think this should be split into two sentences.
```
This documentation is for Deis Workflow version 2. For Deis version 1 documentation, visit...
```

mboersma:
> I may PR against the v1 docs...

Yes, please. This banner will save a lot of confusion.

themes/deis/base.html
jchauncey: I agree I think that would read better.

slack:
Updated, rebased. Will merge when build is done.

PR #72, opened 2016-03-06
Bring documentation in-line with Workflow terminology.

slack:
Use Workflow instead of Deis where it makes sense. Updated all links and page locations.

Of note, I also pulled out most of the left-side links because it was very easy to get yourself into deis.io hierarchy and back into v1 docs (especially via Overview). Helped a bit by #71 

Will be a bit of a rebase mess, but I'm happy to update this PR after #68 and #70 are merged.


arschles:
@slack LGTM, but I'll hold off on labelling since you said you want to wait until #68 and #70 are merged. Let me know when you'd like a final review

slack:
Just rebased now that the others are in. We still good with this?

arschles:
LGTM3

slack:
Good enough for me. :dancer:

PR #91, opened 2016-03-16
doc(managing): Remove managing disk usage page

krancour:
afaik (and someone correct me, _please_, if I have this wrong), the kubelet handles "garbage collection" for dead containers and stale/unused images.  In contrast to Deis v1.x, I don't think there's anything to really say on the topic of managing disk usage.


arschles:
Related question - does it make sense to add docs that explain how to apply restrictions on the disk resource for the minio pod? Although it uses ephemeral disk, it could get large.

krancour:
@arschles I'm wondering if that's an advanced enough topic that it belongs in that component's own docs?  idk

arschles:
@krancour good idea. I've created https://github.com/deis/minio/issues/88 to track

PR #92, opened 2016-03-16
doc(install): Update install directions for renamed chart

krancour:


src/installing-workflow/installing-deis-workflow.md
```diff
@@ -43,9 +43,11 @@ $ helm repo add deis https://github.com/deis/charts
 Now that you have Helm installed and have added the Deis Chart Repository, install Workflow by running:
 
 ```
-$ helm fetch deis/deis                 # fetches the chart into a local workspace
-$ helm generate -x manifests deis      # generates various secrets
-$ helm install deis                    # injects resources into your cluster
+$ helm fetch deis/workflow             # fetches the chart into a local
+                                       # workspace
```
krancour: Note the new manual line break in the comment and continuation on the next line has everything to do with the width of the code windows in the formatted docs.

mboersma:
Should we be referring to the workflow-dev chart for now? The workflow chart houses the alpha1 release and shouldn't be recommended until the v2 release actually lives there I think. OTOH, there may be a workflow-beta1 chart soon that would be more appropriate.

bacongobbler:
I think workflow will eventually become the beta1 release target, AFAIK. I think we should stick to referring to the official "workflow" chart by default, and then "workflow-dev" for bleeding edge dev versions.

krancour:
I'm with @bacongobbler.  I don't think the official docs should be showing how to install the edge.  They should be showing how to install something stable-- even if that something stable is a little old by our fast-paced standards.  As soon as the "workflow" chart is changed to reference the cut beta, the docs will already be right.  :smile:

mboersma:
> As soon as the "workflow" chart is changed to reference the cut beta..

That's what I'm saying should not be done, although it's open for discussion. What's the upgrade path if we always point to the same chart? How would I install Deis v2 beta1 after it's released and Deis commits rc1 or v2 release into the same chart?

arschles:
Chiming in here - I was under the impression we were going to cut a `deis/workflow-beta1`, which would mean we're pointing to a different chart from beta to RC/canary. I assume that depends on the mechanics of helm.

sgoings:
@arschles and @mboersma's comments are what we all decided to do:

> cut a deis/workflow-beta1

krancour:
Ok, gents.  I've fixed this now to mention the `workflow-beta1` chart.

PR #100, opened 2016-03-17
docs(installing-deis): add docs on database config

### LINKED ISSUE #95: Add complete documentation for database object storage configuration
[The DB config section](https://github.com/deis/workflow/blob/master/src/installing-workflow/configuring-object-storage.md#deisdatabase) should specify exactly how to configure it for minio, S3 and Google Cloud Storage (in compatibility mode)

This issue was created in response to the discussion in https://github.com/deis/postgres/issues/72.

bacongobbler:
also fix up dead links to the deis-dev chart

closes #95

### DIFF
--- a/src/installing-workflow/configuring-object-storage.md
+++ b/src/installing-workflow/configuring-object-storage.md
@@ -21,7 +21,7 @@ Every Deis component that relies on object storage relies on the following two i
 - One or more environment variables with host and port to describe where the object storage system is
 - One or more files to provide access credentials for the object storage system.
 	- We suggest storing these values in [Kubernetes secrets](http://kubernetes.io/v1.1/docs/user-guide/secrets.html) and mounting them as volumes to each pod
-	- See [the deis-dev chart](https://github.com/deis/charts/tree/master/deis-dev) for examples of using and mounting secrets.
+	- See [the workflow-dev chart](https://github.com/deis/charts/tree/master/workflow-dev) for examples of using and mounting secrets.
 
 The subsections herein explain how to configure these two inputs for each applicable component.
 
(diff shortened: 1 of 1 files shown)


src/installing-workflow/configuring-object-storage.md
```diff
@@ -111,4 +111,38 @@ The database looks for a `DATABASE_STORAGE` environment variable, which it then
...
+             - name: DATABASE_STORAGE
+               value: {{default "minio" .storage}}
++            - name: DEIS_MINIO_SERVICE_HOST
++              value: storage.googleapis.com
++            - name: DEIS_MINIO_SERVICE_PORT
++              value: "443"
+           readinessProbe:
+             exec:
+               command:
+```
```
arschles: Can this be provided in a plain text format, instead of a diff? Also, since it shows the template syntax, I think this needs to mention that it needs to be installed with `helm generate && helm install`.

bacongobbler: I would love to have it in a plain text format, but I couldn't think of the proper way to explain it to a 5-year old without getting too technical/confusing for the user.

Agreed that we need to call out that this change needs to be done before the user runs `helm generate`.

arschles: Thoughts on something like the following?


Add these values to your spec.template.spec.containers[0].env section:

```yaml
- name: DEIS_MINIO_SERVICE_HOST
  value: storage.googleapis.com
- name: DEIS_MINIO_SERVICE_PORT
  value: "443"
```

bacongobbler: Excellent. I've added something similar to that in 4dbfa77. Does that work for you?

arschles: perfect. the only change I'd make is to make spec.template.spec.containers[0].env into `spec.template.spec.containers[0].env`. I'll also comment on the commit.

arschles:
:+1:

arschles:

Issue #101, opened 2016-03-17
Add information on the right secret to use to the object storage docs


arschles:
Each `Credentials` section mentions the paths that the component looks in, but does not explain practically how to create those paths. Each section should reference the secret (i.e. `minio-user`) to change in the chart in order to make the necessary changes.

PR #104, opened 2016-03-17
doc(object-storage): add sections on configuring secrets for the helm chart

arschles:
Fixes https://github.com/deis/workflow/issues/101

### DIFF
--- a/src/installing-workflow/configuring-object-storage.md
+++ b/src/installing-workflow/configuring-object-storage.md
@@ -52,6 +52,10 @@ The builder reads credentials from the below locations on the filesystem.
 - Key: `/var/run/secrets/object/store/access-key-id`
 - Secret `/var/run/secrets/object/store/access-key-secret`
 
+### Helm Chart
+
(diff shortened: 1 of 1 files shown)


src/installing-workflow/configuring-object-storage.md
```diff
@@ -51,6 +51,10 @@ The builder reads credentials from the below locations on the filesystem.
 - Key: `/var/run/secrets/object/store/access-key-id`
 - Secret `/var/run/secrets/object/store/access-key-secret`
 
+### Helm Chart
+
+If you are using the [Helm Chart for Workflow][helm-chart], put your credentials in the [`minio-user` secret][minio-user-secret] (under `access-key-id` and `access-secret-key`) before you `helm install`. For more information, see the [installation instructions][helm-install] for more details on using Helm.
```
bacongobbler: we should make a note here that the access key and secret **must** be `base64` encoded.

bacongobbler: not sure if this is totally necessary though... Seems like this information would be repeated in multiple locations.

arschles: I think it's worth putting it everywhere. I've made the mistake of _not_ encoding it many, many times

arschles: Added the notes about base64 encoding in bc58dc4e67ae9b110a1be96833c26a71489e10c3

PR #105, opened 2016-03-17
doc(object-storage): add note on beta status

arschles:


jchauncey:
@arschles this is g2g

arschles:
thanks @jchauncey

PR #120, opened 2016-03-21
feat(themes): add static version of deis.com theme

mboersma:
This is a quick-and-dirty port of the theme used at https://deis.com/ over to a MkDocs-compatible [custom theme](http://www.mkdocs.org/user-guide/custom-themes/). There are probably all sorts of things wrong with this, but everyone hates the [current styling](http://docs-v2.readthedocs.org/en/latest/), so if this is preferable we could spend a couple more hours to polish it up.

screen shot 2016-03-21 at 9 18 17 am


bacongobbler:
This looks awesome! We're definitely going to need to tweak the CSS though so we get a navbar, better styling for code blocks etc. but this is a great start.

mboersma:
This was a quick weekend hack. I'll leave the CSS-slinging to the professionals who are now on this task.

PR #132, opened 2016-04-01
docs(reference-guide): update deis logs response

bacongobbler:
port of deis/deis#5004

to be tested with deis/controller#592


src/reference-guide/controller-api/v2.0.md
```diff
@@ -337,7 +337,8 @@ DEIS_API_VERSION: 2.0
 DEIS_PLATFORM_VERSION: 1.12.1
```
mboersma: This version is actually `2.0.0-dev`, but that's out of scope for this PR. I'll make an issue for that.

bacongobbler: :+1:

bacongobbler:
Closing since this is not true; see #202

PR #139, opened 2016-04-07
feat(customizing-workflow): add component tuning

bacongobbler:
In v1 we had a section dedicated to customizations that users could
add to each component. This brings back that documentation back,
adding only controller and database-specific customizations to the
mix with others to be added as necessary.

addresses #75

Issue #79, opened 2016-03-15
Meta issue-- production deployments


krancour:
This page is currently a placeholder: https://github.com/deis/docs-v2/blob/master/src/managing-workflow/production-deployments.md

I want to start a simple checklist here of recommendations we _must_ document for anyone wishing to run Workflow in production.  I invite others to edit this.

- [ ] Must: Use off-cluster object storage (or else everything is ephemeral!).  Should include a link to object storage configuration page.
- [ ] Recommended: Use off-cluster Postgres?
- [ ] _Highly_ recommended: Secure the Workflow cluster using an SSL wildcard cert
- [ ] Recommended: Generate a unique dhparam for the router (to further harden Diffie Hellman key exchange)

Do we want to make any recommendations re: the k8s clusters themselves?  Or is this not our concern?  Example: speaking to @slack today, I believe we intend in our dog-fooding cluster to have the k8s worker nodes live in private subnets.  I don't know of any pre-baked zero-to-k8s solutions that account for that... so is it worth mentioning?  Or maybe is it worth having a whole _separate_ checklist for things like this and articulating that "A Deis Workflow cluster is only as good as the Kubernetes cluster you run it on.  Here are some considerations..."

Discuss... discuss...

bacongobbler:
for reference, I PR'd #141 that tackles a few of these concepts by porting over most of the v1 documentation. So far to address your points:

> Use off-cluster object storage (or else everything is ephemeral!). Should include a link to object storage configuration page.

Done, however we should really have another document that explains how to use RDS and S3.

> Recommended: Use off-cluster Postgres?

Related to point 1 ^^

> Highly recommended: Secure the Workflow cluster using an SSL wildcard cert

That's been ported over :)

> Recommended: Generate a unique dhparam for the router (to further harden Diffie Hellman key exchange)

That has not been documented, though if you have any input or could point to some documentation on how one would go and generate their own unique dhparam, that would be greatly appreciated.

As for recommendations on how the network topology should work, I think we need to "sip our own champagne" first and figure out how we want to handle that internally before recommending topology best practices.

jchauncey:
moving to rc1

sstarcher:
#199 would help out the s3 folks who are on aws

bacongobbler:
Once that's implemented, yes we can add that to this doc.

arschles:
https://github.com/deis/workflow/issues/180 would also help AWS folks

jgmize:
[...]

PR #141, opened 2016-04-07
feat(managing-workflow): add back production deployments

bacongobbler:


bacongobbler:
relies on #139 before this can go green, since this doc references that PR

bacongobbler:
refs #79

src/managing-workflow/production-deployments.md
```diff
@@ -1,3 +1,57 @@
 # Production Deployments
 
-TODO (bacongobbler): rewrite for v2
+Many Workflow users are running Workflow quite successfully in production. When readying a Workflow
```
mboersma: I assume this is a forward-looking statement. Still, I would drop the first sentence completely until such time as it becomes accurate.

src/managing-workflow/production-deployments.md
```diff
@@ -1,3 +1,57 @@
...
+users can consider enabling a firewall on the CoreOS hosts as well as the router component.
+
+See [Security Considerations][] for details.
+
+
+## Change Registration Mode
+
+Changing the registration process is highly recommended in production. By default, registrations
+for a new cluster are open to anyone with the proper URL. Once the admin user has registered with a
+new cluster, it is recommended that you either turn off registrations entirely or enable the admin
```
mboersma: I would say instead:
> it is recommended either to disable registrations or to enable the admin-only registration feature.

Specifically, I prefer avoiding "you" when possible, and "admin-only" should contain a hyphen.

src/managing-workflow/production-deployments.md
bacongobbler: Agreed. This is definitely a little forward-thinking. It was a copy-over from the old v1 documentation.

mboersma:
This has a broken hyperlink:

> Error: The page "managing-workflow/production-deployments.md" contained a hyperlink to "customizing-workflow/tuning-component-settings.md" which is not listed in the "pages" configuration.

bacongobbler:
@mboersma see https://github.com/deis/workflow/pull/141#issuecomment-207121700 above :)

src/managing-workflow/production-deployments.md
```diff
@@ -1,3 +1,57 @@
...
-TODO (bacongobbler): rewrite for v2
+When readying a Workflow deployment for production workloads, there are some additional (but
+optional) recommendations.
+
+
+## Running Workflow without Minio
+
+Workflow makes use of [Minio][] to provide persistent storage for the [Registry][], [Database][],
+and [Logger][] components. The additional operational complexity of Minio is tolerated because of
+the need for persistent storage for platform high availability.
```
mboersma: I don't think it's accurate to describe Minio as complex. Also "high availability" implies that deis/minio is resilient in ways it unfortunately isn't.

src/managing-workflow/production-deployments.md
```diff
@@ -1,3 +1,57 @@
...
+
+Alternatively, persistent storage can be achieved by running an external S3-compatible blob store,
```
mboersma: Really this should be encouraged rather than presented as just an alternative. Maybe this could just remove `Alternatively,`?

src/managing-workflow/production-deployments.md
```diff
@@ -1,3 +1,57 @@
 # Production Deployments
 
-TODO (bacongobbler): rewrite for v2
+When readying a Workflow deployment for production workloads, there are some additional (but
```
mboersma: Since they are recommendations, you could omit "(but optional)" as it's redundant.

src/managing-workflow/production-deployments.md
mboersma: Honestly, maybe just use "storage" instead of "persistent storage" throughout this paragraph, and drop the "high availability" phrase.

mboersma:
```
DEBUG   -  Building page managing-workflow/production-deployments.md 
ERROR   -  Error building page managing-workflow/production-deployments.md 
Error: The page "managing-workflow/production-deployments.md" contained a hyperlink to "customizing-workflow/tuning-component-settings.md" which is not listed in the "pages" configuration.
make: *** [build] Error 1
```

bacongobbler:
All great points. Thanks @mboersma! Most of this is a rip from the old "production deployments" doc but I agree that a lot of it does not make sense now. I've already had to edit the old doc quite a bit, so thank you for picking up on where I've missed! I'll rip those out.

bacongobbler:
and @mboersma see above, you already pointed that out in https://github.com/deis/workflow/pull/141#issuecomment-214466052. I'm still waiting on #139 to be merged before that hyperlink won't break any more.

bacongobbler:
@mboersma all comments have been addressed. Thank you for taking the time to review. :)

PR #143, opened 2016-04-07
feat(managing-deis): lazy pass at upgrading workflow

bacongobbler:


src/managing-workflow/upgrading-workflow.md
```diff
@@ -0,0 +1,11 @@
+# Upgrading Workflow
+
+To upgrade to a newer version of Workflow, run the following:
+
+```
+$ helm uninstall workflow-beta1 -n deis
+$ helm update
+$ helm fetch deis/workflow-beta2
+$ helm generate workflow-beta2
```
mboersma: Maybe do `helm generate -x manifests workflow-beta2` here? Otherwise users may be confused by helm/helm#393.

PR #148, opened 2016-04-09
doc(readme): help readme function as a landing page

slack:
Until we have a more sophisticated "try/get-started" page on deis.com we are sending folks to `deis/workflow`. Updated the copy to direct folks to the next logical step in the setup process.

Fixes engineyard/deis.com#102


slack:
Woo, already lgtmd!

PR #159, opened 2016-04-11
ref(installing-workflow): rename workflow-dev to workflow-beta1

### LINKED ISSUE #129: Make workflow docs point to the workflow-beta3 chart
Currently, some docs (such as [object storage](https://github.com/deis/workflow/blob/master/src/installing-workflow/configuring-object-storage.md)) point to https://github.com/deis/charts/tree/master/workflow-dev, but they should probably point to https://github.com/deis/charts/tree/master/workflow-beta1

bacongobbler:
closes #129


mboersma:
`s/beta1/beta2/g`??

bacongobbler:
Yep. Lemme do that real quick

bacongobbler:
This has already been updated in master so closing this

PR #160, opened 2016-04-12
fix(themes): fix up navigation on readthedocs

bacongobbler:
The version list is not being shown on readthedocs.org because it's
not in the navigation list.

addresses #1 but will keep open as I cannot verify until it's pushed and deployed to readthedocs. No breaking changes occurred locally, however.


mboersma:
I tested this locally, and also removed the `if READTHEDOCS` test, but it's not really going to work until it's deployed to rtfd.org. I can verify this does no harm though.

PR #167, opened 2016-04-17
doc(objectstorage): refactor configuring object storage documentation

slack:
Couple of things I wanted to accomplish here.

* Make the top of the doc goal oriented to actually walk folks through wiring this up.
* Remove as much repeated content as possible like the redundant helm charts sections.
* Expand on the paths and env vars used to help point folks in the right direction if debugging.

Pretty major rewrite, so I wanted to throw this up for feedback. Haven't given it a close read for spelling and grammar.

### DIFF
--- a/src/installing-workflow/configuring-object-storage.md
+++ b/src/installing-workflow/configuring-object-storage.md
@@ -1,140 +1,116 @@
 # Configuring Object Storage
 
-A variety of Deis components rely on an object storage system to do their work. These components are:
+A variety of Deis Workflow components rely on an object storage system to do their work including storing application slugs, Docker images and database logs.
 
-- [builder](https://github.com/deis/builder)
-- [slugbuilder](https://github.com/deis/slugbuilder)
-- [slugrunner](https://github.com/deis/slugrunner)
-- [controller](https://github.com/deis/controller)
-- [registry](https://github.com/deis/registry)
-- [database](https://github.com/deis/postgres)
+Deis Workflow ships with [Minio][minio] by default, which provides in-cluster, ephemeral object storage. This means that _if the Minio server crashes, all data will be lost_. Therefore, **Minio should be used for development or testing only**.
 
-These components are flexible and can work out of the box with almost any system that is compatible with the [S3 API](http://docs.aws.amazon.com/AmazonS3/latest/API/APIRest.html).
+## Configuring off-cluster Object Storage
 
-## Minio
+Every component that relies on object storage uses two inputs for configuration:
 
-Additionally, Deis ships with a [Minio](http://minio.io) [component](https://github.com/deis/minio) by default, which provides in-cluster, ephemeral object storage. This means that _if the Minio server crashes, all data will be lost_. Therefore, **Minio should be used for development or testing only**.
+1. Component-specific environment variables (e.g. `BUILDER_STORAGE` and `REGISTRY_STORAGE`)
+2. Access credentials stored as a Kubernetes secret named `objectstorage-keyfile`
 
-In our beta release, the components listed above are configured by default to automatically use the Minio [service][k8s-service] for object storage.
+The helm chart for Deis Workflow can be easily configured to connect Workflow components to off-cluster object storage. Deis Workflow currently supports Google Compute Storage, Amazon S3 and Azure Blob Storage.
 
-# Configuring the Deis Components
+* **Step 1:** Create storage buckets for each of the Workflow subsystems: builder, registry and database
+    * Note: Depending on your chosen object storage you may need to provide globally unique bucket names.
+    * Note: If you provide credentials with sufficient access to the underlying storage, Workflow components will create the buckets if they do not exist.
+* **Step 2:** If applicable, generate credentials that have write access to the storage buckets created in Step 1
+* **Step 3:** If you haven't already fetched the helm chart, do so with `helm fetch deis/workflow-beta2`
+* **Step 4:** Open the helm chart with `helm edit workflow-beta2` and look for the template file `tpl/generate_params.toml`
+* **Step 5:** Update the `storage` parameter to reference the storage platform you are using: `s3`, `azure`, `gcs`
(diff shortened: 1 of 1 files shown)


src/installing-workflow/configuring-object-storage.md
```diff
@@ -1,140 +1,117 @@
...
 
-The subsections herein explain how to configure these two inputs for each applicable component.
+During the `helm generate` step, Helm creates a Kubernetes secret in the Deis namespace named `objectstorage-keyfile`. The exact structure of the file depends on storage backend specified in `tpl/generate_params.toml`.
 
-## [deis/builder](https://github.com/deis/builder)
+```
+# Set the storage backend
+#
+# Valid values are:
+# - filesystem: Store persistent data on ephemeral disk
```
kmala: Only registry supports the file system and it is present in the chart because initially only registry used to use this secret.So i think we should not keep it still as a valid value.Charts also need to be updated.

slack: Ok. I'll update this PR.

src/installing-workflow/configuring-object-storage.md
```diff
@@ -1,140 +1,116 @@
...
-A variety of Deis components rely on an object storage system to do their work. These components are:
+A variety of Deis Workflow components rely on an object storage system to do their work including storing application slugs, Docker images and database logs.
 
-- [builder](https://github.com/deis/builder)
-- [slugbuilder](https://github.com/deis/slugbuilder)
-- [slugrunner](https://github.com/deis/slugrunner)
-- [controller](https://github.com/deis/controller)
-- [registry](https://github.com/deis/registry)
-- [database](https://github.com/deis/postgres)
+Deis Workflow ships with [Minio](http://minio.io) [component](https://github.com/deis/minio) by default, which provides in-cluster, ephemeral object storage. This means that _if the Minio server crashes, all data will be lost_. Therefore, **Minio should be used for development or testing only**.
```
bacongobbler: "Minio component" seems a little awkward. Would it be better to just link "minio" to `../understanding-deis/components.md#object-storage` instead?

src/installing-workflow/configuring-object-storage.md
```diff
@@ -1,140 +1,116 @@
...
 
-Additionally, Deis ships with a [Minio](http://minio.io) [component](https://github.com/deis/minio) by default, which provides in-cluster, ephemeral object storage. This means that _if the Minio server crashes, all data will be lost_. Therefore, **Minio should be used for development or testing only**.
+1. Component-specific environment variables (e.g. `BUILDER_STORAGE` and `REGISTRY_STORAGE`)
+2. Access credentials stored as a Kubernetes secret named `objectstorage-keyfile`
 
-In our beta release, the components listed above are configured by default to automatically use the Minio [service][k8s-service] for object storage.
+The helm chart for Deis Workflow can be easily configured to connect Workflow components to off-cluster object storage. Deis Workflow currently supports Google Compute Storage, Amazon S3 and Azure Blob Storage.
 
-# Configuring the Deis Components
+* **Step 1:** Create buckets for each of the workflow subsystems: builder, registry and database
```
bacongobbler: not necessary. The components will create the buckets if they do not exist. :)

src/installing-workflow/configuring-object-storage.md
```diff
@@ -1,140 +1,116 @@
...
+* **Step 2:** Create user credentials that have access to the storage buckets from step 1
```
bacongobbler: considering step 1, we should just say to "create user credentials"

src/installing-workflow/configuring-object-storage.md
slack: Ok. Trying to prevent folks from picking crappy bucket names like "deis" and expecting them to work and the components will explode in a non-obvious way.

Do you think a section on names, buckets and such would be more useful?

src/installing-workflow/configuring-object-storage.md
slack: I can rephrase. This was the existing copy.

src/installing-workflow/configuring-object-storage.md
slack: This was motivated by the encouraged separation of concerns. IAM creds for the writers, that may or may not have permission to manipulate buckets.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Action": [
                "s3:*"
            ],
            "Effect": "Allow",
            "Resource": [
                "arn:aws:s3:::bucket-registry",
                "arn:aws:s3:::bucket-registry/*",
                "arn:aws:s3:::bucket-database",
                "arn:aws:s3:::bucket-database/*",
                "arn:aws:s3:::bucket-builder",
                "arn:aws:s3:::bucket-builder/*"
            ]
        }
    ]
}
```

src/installing-workflow/configuring-object-storage.md
bacongobbler: perhaps just a note in the `helm edit workflow-beta2` section when they're changing storage parameters would be good.

src/installing-workflow/configuring-object-storage.md
```diff
@@ -1,140 +1,116 @@
...
+### [deis/registry](https://github.com/deis/registry)
 
-## [deis/slugrunner](https://github.com/deis/slugrunner)
+The registry looks for a `REGISTRY_STORAGE` environment variable which it then uses as a key to look up the object storage location and authentication information.
 
-The slugrunner is configured and launched by the controller inside a Deis cluster, so this section only applies if you intend to run it as a standalone component.
+The registry reads credential information by reading `/var/run/secrets/deis/registry/creds/objectstorage-keyfile`.
 
-### Environment Variables
+This is the file location for the `objectstorage-keyfile` secret on the Pod files.sytem.
```
arschles: typo: `files.sytem` => `filesystem`

arschles:
@slack other than [the typo](https://github.com/deis/workflow/pull/167#discussion_r60138319) this LGTM

arschles:
Closes #169

PR #183, opened 2016-04-22
docs(using-docker-images): add --no-remote

rimusz:


src/using-workflow/using-docker-images.md
```diff
@@ -33,13 +33,9 @@ In order to deploy Docker images, they must conform to the following requirement
...
 
     $ mkdir -p /tmp/example-go && cd /tmp/example-go
-    $ deis create
+    $ deis create example-go --no-remote
     Creating application... done, created example-go
 
-!!! note
-    The `deis` client uses the name of the current directory as the
-    default app name.
-
```
arschles: @rimusz seems like it'd be good to keep this here, so people know that they don't _have_ to specify the app name if they don't want. What do you think of the below?

```
!!! note
    The `deis` client uses the name of the current directory as the default app name if you don't specify it, as above
```

src/using-workflow/using-docker-images.md
```diff
@@ -33,13 +33,9 @@ In order to deploy Docker images, they must conform to the following requirement
 Use `deis create` to create an application on the [controller][].
 
     $ mkdir -p /tmp/example-go && cd /tmp/example-go
-    $ deis create
+    $ deis create example-go --no-remote
     Creating application... done, created example-go
 
-!!! note
```
bacongobbler: I'd keep this note in since the rest of the commands don't use `--app` to specify the application name.

src/using-workflow/using-docker-images.md
rimusz: @arches that does no apply for apps with --no-remote:
```
 $ deis create --no-remote example-go
Creating Application... done, created example-go
remote available at ssh://git@deis.130.211.97.132.nip.io:2222/example-go.git
-----
$ deis create --no-remote
Creating Application... done, created bubbly-quantity
remote available at ssh://git@deis.130.211.97.132.nip.io:2222/bubbly-quantity.git
```

bacongobbler: that's because you're in `/tmp/example-go`, so the name is inferred from the subdirectory. If you were in `tmp/foo` you'd get a 404 NOT FOUND error because the app name `foo` was not found.

rimusz: @bacongobbler what `/tmp/example-go` are you talking about?

bacongobbler: line 35.

rimusz: still the same:
```
$ mkdir -p /tmp/example-go && cd /tmp/example-go
$ deis create --no-remote
Creating Application... done, created wonder-elephant
remote available at ssh://git@deis.130.211.97.132.nip.io:2222/wonder-elephant.git
```

bacongobbler: I'm referring to the other commands below like `deis pull`. For example, try the following:

```
><> mkdir /tmp/foo
><> cd /tmp/foo/
><> deis create example-go --no-remote
Creating Application... done, created example-go
remote available at ssh://git@deis.10.245.1.3.xip.io:2222/example-go.git
><> deis pull deis/example-go
Creating build... Error: 
404 Not Found
detail: Not found.
```

That's why I'm saying that this note is important. Not to `deis create`, but to every other command in the CLI that targets an existing application.

bacongobbler: Perhaps a better note to add onto @arschles's would be

```
!!! note
    For all commands except for `deis create`, the `deis` client uses the name of the current directory as the app name if you don't specify it explicitly with `--app`.
```

arschles: works for me!

rimusz: the last one makes more sense, updating PR

Issue #185, opened 2016-04-25
Add workflow-cli to the "What's in a release" section of releasing.md


arschles:
It should be under the "Components" section

PR #187, opened 2016-04-25
doc(releasing.md): add workflow-cli to the list of components

arschles:
Fixes https://github.com/deis/workflow/issues/185

Also removes errant tag & git SHA information from workflow-manager


bacongobbler:
thanks! I was wondering what that git SHA was doing.

PR #190, opened 2016-04-25
doc(releasing.md): merge helm documentation update step into the helm chart update step

arschles:
It's easier to update and patch the charts repo with the entire new helm chart, including `README.md` and `Chart.yaml` updates, at once


bacongobbler:
ooh good thought. That would certainly be nice.

PR #201, opened 2016-04-26
feat(registry): add private registry documentation

helgi:


src/using-workflow/using-docker-images.md
```diff
@@ -58,12 +58,29 @@ Use `deis scale cmd=3` to increase `cmd` processes to 3, for example. Scaling a
 process type directly changes the number of [Containers][container]
 running that process.
 
-!!! attention
-    Support for Docker registry authentication is coming soon.
+## Private Registry
 
+In order to deploy Docker images from a private repository on to Deis Workflow the following steps
```
mboersma: > To deploy Docker images from a private registry, take the following steps:

* Gather the username and password for the registry, such as a [Quay.io Robot Accounts][] or [GCR.io Long Lived Token][]

src/using-workflow/using-docker-images.md
```diff
@@ -58,12 +58,29 @@ Use `deis scale cmd=3` to increase `cmd` processes to 3, for example. Scaling a
...
+need to be taken before deployment:
+
+* Gather username and password for the repository, such as from [Quay.io Robot Accounts][] or [GCR.io Long Lived Token][]
+* Run `deis registry:set username=<the-user> password=<secret> -a <application-name>`
+* Now do `deis pull` as normally, against an image in a private repository
+
+When using a [GCR.io Long Lived Token][] the JSON blob will have to be compacted first using a tool like [jq][]
```
mboersma: This sentence should end with a period.

src/using-workflow/using-docker-images.md
```diff
@@ -58,12 +58,29 @@ Use `deis scale cmd=3` to increase `cmd` processes to 3, for example. Scaling a
...
+
+`cat serviceaccount.json | jq -c .` will give the desired result
+
+!!! note
```
mboersma: Maybe use `**NOTE:**` instead here so it renders as **NOTE:** ?

src/using-workflow/using-docker-images.md
```diff
@@ -58,12 +58,29 @@ Use `deis scale cmd=3` to increase `cmd` processes to 3, for example. Scaling a
...
-!!! attention
-    Support for Docker registry authentication is coming soon.
+## Private Registry
 
+In order to deploy Docker images from a private repository on to Deis Workflow the following steps
+need to be taken before deployment:
+
+* Gather username and password for the repository, such as from [Quay.io Robot Accounts][] or [GCR.io Long Lived Token][]
+* Run `deis registry:set username=<the-user> password=<secret> -a <application-name>`
+* Now do `deis pull` as normally, against an image in a private repository
```
mboersma: "as normal"

src/using-workflow/using-docker-images.md
helgi: Fair, was only following what the other `note` things in this doc do

krancour:
I'd be careful about using registry and repository interchangeably.  It might be more precise to indicate that this can be used for private registries (e.g. if you host your own) _or_ private repositories _within_ a public registry like Dockerhub or quay.io.

On a separate note, there's mention of compacting the JSON blob for a long-lived token, but there's no mention of how that can be applied.  Is the compacted token just substituted as the value of the `--password` flag?  Or has it got its own dedicated flag?

helgi:
It's in the password, I can update that

src/using-workflow/using-docker-images.md
```diff
@@ -58,12 +58,30 @@ Use `deis scale cmd=3` to increase `cmd` processes to 3, for example. Scaling a
 process type directly changes the number of [Containers][container]
 running that process.
 
-!!! attention
-    Support for Docker registry authentication is coming soon.
+## Private Registry
 
+To deploy Docker images from a private registry (either from self-hosted registry or one of the public ones),
```
mboersma: minor: should say "from **a** self-hosted registry"

src/using-workflow/using-docker-images.md
```diff
@@ -58,12 +58,30 @@ Use `deis scale cmd=3` to increase `cmd` processes to 3, for example. Scaling a
 process type directly changes the number of [Containers][container]
 running that process.
 
-!!! attention
-    Support for Docker registry authentication is coming soon.
+## Private Registry
 
+To deploy Docker images from a private registry (either from a self-hosted registry or one of the public ones),
+take the following steps:
```
krancour: Suggested change:

"To deploy Docker images from a private registry or even from a private repository within a public registry, take the following steps:"

src/using-workflow/using-docker-images.md
```diff
@@ -58,12 +58,29 @@ Use `deis scale cmd=3` to increase `cmd` processes to 3, for example. Scaling a
...
 
-!!! attention
-    Support for Docker registry authentication is coming soon.
+## Private Registry
 
+To deploy Docker images from a private registry or even from a private repository within a public registry, take the following steps:
+
+* Gather the username and password for the registry, such as a [Quay.io Robot Accounts][] or [GCR.io Long Lived Token][]
+* Run `deis registry:set username=<the-user> password=<secret> -a <application-name>`
+* Now do `deis pull` as normal, against an image in the private registry
```
krancour: Would "perform" sound better than "do"?

krancour: Total nit... non-blocking.  LGTM2

rimusz:
👍

skevy:
Are there any thoughts on how to make an entire deis cluster have access to a private registry? Or is that not in the cards/very technically difficult.

helgi:
Right now it is only per application. If we ever do https://github.com/deis/controller/issues/383 then registry data may get included in that as well

skevy:
That would be great. Basically I don't want to have members of my team have to ask me for a registry key :)

helgi:
My recommendation would be to make a read-only robot account credentials available to them, in the short term

skevy:
👍

PR #239, opened 2016-05-18
docs(using-workflow): clarify what username and password reflect

bacongobbler:
some users were getting confused what exactly should be in the username/password fields. Clarifying
that these are identical to what's used in `docker login` should help clear up the air.

### DIFF
--- a/src/using-workflow/using-docker-images.md
+++ b/src/using-workflow/using-docker-images.md
@@ -37,7 +37,7 @@ Use `deis create` to create an application on the [controller][].
     Creating application... done, created example-go
 
 !!! note
-    For all commands except for `deis create`, the `deis` client uses the name of the current directory 
+    For all commands except for `deis create`, the `deis` client uses the name of the current directory
     as the app name if you don't specify it explicitly with `--app`.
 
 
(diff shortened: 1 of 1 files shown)


src/using-workflow/using-docker-images.md
```diff
@@ -60,26 +60,33 @@ running that process.
...
+
+* Gather the username and password for the registry, such as a [Quay.io Robot Account][] or a [GCR.io Long Lived Token][]
 * Run `deis registry:set username=<the-user> password=<secret> -a <application-name>`
 * Now perform `deis pull` as normal, against an image in the private registry
 
-When using a [GCR.io Long Lived Token][] the JSON blob will have to be compacted first using a tool like [jq][]
-and then used in the password field in `deis registry:set`.
+When using a [GCR.io Long Lived Token][], the JSON blob will have to be compacted first using a
+tool like [jq][] and then used in the password field in `deis registry:set`. For the username, use
+`_token_json`. For example:
```
helgi: `_json_key` ?

bacongobbler: derp. thank you

PR #259, opened 2016-05-24
docs(upgrading-workflow): describe "keeper" upgrade behavior

mboersma:


src/managing-workflow/upgrading-workflow.md
```diff
@@ -9,3 +9,24 @@ $ helmc fetch deis/workflow-beta4
...
+and for the router as keepers. This keeps the external `LoadBalancer` intact so that routing
+and DNS are preserved while reinstalling Workflow.
+
+To remove Workflow completely from a Kubernetes cluster, run the following:
+
+```
+$ helmc uninstall -n deis -y workflow-dev
+$ kubectl delete ns deis  # Be sure you want to delete the
+                          # Namespace and all its contents!
+```
```
arschles: Is it worth saying why someone would want to use this "keeper" functionality? I'm assuming it's primarily for upgrading an existing cluster? Anyway, not blocking for this PR

mboersma: I think the second paragraph describes why Workflow uses "keeper" (yes, for upgrade), but let me know how I can clarify. The general `helm-keep` mechanism is described at <http://helm-classic.readthedocs.io/en/latest/awesome/#keeper-manifests>. Would it help to link to that doc from here?

arschles: @mboersma I _think_ it would help, yes, but again not blocking for this PR

PR #268, opened 2016-05-25
docs(managing-workflow): Add scaling the routers to the production deployment notes.

mattk42:


kmala:
There are few issues which you might face when you scale the router and hence it is not included in the docs.
https://github.com/deis/router/issues/152

mattk42:
Ah, good to know. In my case I would rather have to wait that extra 10s when a new routable service is added than risk the potential > 10s outage for all services in the event of a single router pod failing for whatever reason. But I could see why you may not want that in the docs just yet.

src/managing-workflow/production-deployments.md
```diff
@@ -46,6 +46,12 @@ Please see the following documentation: [Customizing Controller][]
 Using TLS to encrypt traffic (including Workflow client traffic, such as login credentials) is
 crucial. See [Platform SSL][] for the platform.
 
+## Scale Routers
+
+If all router pods in your cluster become unavailable then you will be unable to access the workflow API or
+any deployed applications. To reduce the potential of this happening it is recommended that you scale the
+deis-router ReplicaController to run more than one replica controller. This can be accomplished by running
```
bacongobbler: FYI you're not scaling replica controllers. The replica controller manages pods, so it'd be better to say

>  To reduce the potential of this happening, it is recommended that you scale the Router's ReplicaController to run more than one pod

Though I agree with keerthan on how we're only recommending one replica for v2 at this time.

krancour:
I agree with @mattk42's reasoning. I think as long as the documentation notes the caveat or two we know of WRT scaling router, then it's still worth showing how it's done.

jchauncey:
@mattk42 @krancour @kmala what do we need to do to get this merged?

bacongobbler:
GIven that this was PR'd before we moved to Deployments, any references to Replication Controllers will probably need to be updated.

ged15:
Correct me if I'm wrong, but wouldn't having only 1 replica available, mean downtime of the router during an upgrade? A small downtime window, but there nonetheless.

ged15:
Could we also have a note on what the trade-offs of increasing no. of replicas is? Also will the number of replicas remain after a `helm upgrade`?

bacongobbler:
The only downside is that the routers aren't watching the kubernetes API; they operate on a poll loop, updating config every 10 seconds. By design they can be out of sync.

bacongobbler:

mboersma:
@mattk42 could you amend the commit messages themselves to be something like:
```
docs(managing-workflow): add router scaling to production deployment
```

Otherwise these commits get dropped from the CHANGELOG generation.

If you're too busy, we can tweak the messages and open a new PR. Thanks again!

mattk42:
@mboersma All set, sorry about that.

mboersma:
Jenkins, add to whitelist.

PR #290, opened 2016-06-07
docs(quickstart/deploy-an-app): explain how to add user ssh key

rimusz:


bacongobbler:
this was removed from the quickstart because it uses `deis pull`, however I can understand the need to do this. @slack, thoughts on adding this here?

slack:
Yeah. I switched the quickstart to use `deis pull` so we didn't have to _also_ explain checking out an example app in the `git push` flow.

This PR does bring up a good point in that I _didn't_ bring SSH key information into the buildpack and dockerfile sections of the docs. That is probably where this should live. If you wanna PR this into those pages that would be :+1:.

Also make sure that you use relative links to other doc pages so we can more easily update the location this stuff is being hosted. 👯

rimusz:
closing this one, opened #295 for buildpack and dockerfile sections of the docs

PR #294, opened 2016-06-07
doc(release-checklist): Improve instructions pertaining to previous release

krancour:
This adds documentation pertaining to exporting and usage of the `WORKFLOW_PREV_RELEASE ` environment variable (previously `WORKFLOW_OLD_RELEASE `).


krancour:
I've removed the two LGTMs I had. I was forced to rebase this and that rebase had me re-applying my change atop some inaccurate language. I corrected those inaccuracies, but it means this rebased change is probably now worth a re-review.

mboersma:
@krancour this still makes sense to me.

PR #293, opened 2016-06-07
doc(styles): theme the docs to match deis.com

flynnduism:
This PR updates the Documentation theme (using MK Docs) to follow the style and brand of deis.com.

docs

- [x] fix the sidebar menu states
- [x] clean up old theme JS
- [x] minify assets
- [x] browser test

_Note:_ changes to the makefile compile the .com and the docs sass files into a single `styles.css` - but doing so makes sass a requirement. Perhaps I should omit this and only commit compiled css.

PR #302, opened 2016-06-09
doc(styles): theme the docs to match deis.com

slack:
Replaces https://github.com/deis/workflow/pull/293


arschles:
2 things I've noticed so far:

1. On my 15" Macbook Pro Retina screen, when I scroll all the way to the bottom, the footer overlaps the content. Not sure if that's expected behavior?
2. Header links that go to other locations on `deis.com` seem to have hard-coded `deis.com`. Can they be relative links instead?

slack:
Links need to be hard-coded to deis.com since this will live on rtd _and_ deis.com for a period of time.

PR #305, opened 2016-06-09
chore(rtd): set canonical url for documentation to deis.com

slack:
Must be merged before https://github.com/engineyard/deis.com/pull/158

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -3,6 +3,7 @@
 # names
 #
 site_name: Deis Workflow Documentation
+site_url: https://deis.com/docs/workflow
 pages:
   - Home: index.md
   - Quick Start:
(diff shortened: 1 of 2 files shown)


slack:
I'm not 💯  on what rtd will do with this. I'm not sure how to deploy a test branch! So this may need to be reverted.

slack:
Just for a bit more context here. Two goals:
* This change should cause fully qualified links to appear in `sitemap.xml` for all pages in the documentation site. This will be referenced by `deis.com/robots.txt` after merge.
* Additionally, this will add a `meta canonical` tag that points to the corresponding `deis.com/docs/workflow` page location. This helps search engines de-duplicate links which should keep us from avoiding search ranking penalties for what would otherwise appear as duplicate content.

flynnduism:
`LGTM` (I can't seem to add a label for some reason)

Would it also be worth referencing the `sitemap.xml` in the <head> for crawlers to find? i.e.
`<link href="sitemap.xml" rel="feed" type="application/rss+xml" title="Sitemap" />`

slack:
I have a PR which adds the workflow docs sitemap via `deis.com/robots.txt`. I think that should cover us.

slack:
Woop, worked perfectly!

PR #315, opened 2016-06-13
docs(submitting-a-pr): add "infra project" exception to LGTM rule

mboersma:
Much discussion around streamlining code review and approval led to this suggestion.

Unlike #314, I don't think there was total consensus around this change. So let's use this PR to discuss further if needed.

ping @jchauncey

### DIFF
--- a/src/contributing/submitting-a-pull-request.md
+++ b/src/contributing/submitting-a-pull-request.md
@@ -93,6 +93,8 @@ If the PR is from a Deis maintainer, then he or she should be the one to close i
 
 An exception to this is when an errant commit needs to be reverted urgently. If necessary, a PR that only reverts a previous commit can be merged without waiting for LGTM approval.
 
+Deis projects that are considered "infrastructure" may optionally choose to require only one maintainer LGTM before merging code. Infrastructure projects contain CI/CD resources or support tooling, not end-user code.
+
 [go]: http://golang.org/
 [glide]: https://github.com/Masterminds/glide
 [flake8]: https://pypi.python.org/pypi/flake8/


sgoings:
I wanted to shed a little light on the problem before delving into another possible solution...

The problem domain as I remember it was that we have infrastructure projects that aren't as well known as some of our productized components (router, controller, etc.). Infrastructure projects also tend to pop up faster than product projects.

For a specific example, the jenkins-jobs repository. I believe @vdice, @jchauncey, and myself are really the only ones that know "what works and what doesn't" in that repo (although even that statement isn't entirely true either).

A potential solution that's been on my mind for this problem (projects that are hard to get LGTMs for) is maybe we introduce the idea of a project "maturity" index where a prototype project might follow the 1 LGTM rule but a "production" project (one of our workflow components) might follow the 2 (or more! gasp!) LGTM rule.

sgoings:
> I think that keeps the distinction stupidly simple for everyone, which I would prefer, rather than carving out an exemption for what at this point appears to be just jenkins-jobs.

This might be one of the reasons why we differ in direction; I'd like to see jenkins-jobs still remain public, as we've made kind of a big splash already in the community about holding our jenkins job configurations available for all...

bacongobbler:
So in terms of what we consider "infrastructure code", does that include the example apps as well since they are part of the CI/CD pipeline? From what I see, "infrastructure code" translates to these repos:

 - jenkins-jobs
 - k8s-claimer
 - mention-bot
 - deisrel
 - seed-repo
 - chart-mate
 - rigger (v1)
 - all example-* apps

I'm +1 on keeping the infrastructure code public because we definitely want to open source our infrastructure. Having repos for deis.io, ci.deis.io and others in the public eye allows community members to help contribute to them, as well as share knowledge on how we run the project.

> My counterpoint to that is that we should focus on decreasing the bus factor for that repo then, not hold it to a lower standard of review.

At the same time, I agree with this statement. Personally I haven't seem much of an issue with the PR backlog as much with any other project, so maybe it's just a matter of getting more devs to work on these projects so they can consistently get more reviews from other sources.

mboersma:
> does that include the example apps as well?

I would say no, simply because they are expressly user-facing. We intend for new users to `git clone` and work with them as an introduction to pushing their apps into Workflow. So the code should be clean, reviewed, and presentable.

Additionally, there may be a distinction here between "critical" and "optional" infrastructure code. That is, if merging broken code into the project could cause problems for CI or for developers who expect the master branch to work, it should be held to a high review standard. Breaking `example-dockerfile-http`, just for...um...[example](https://github.com/deis/example-dockerfile-http/pull/12), sadly can cause CI to fail and cost lots of developer time, so I think 1-LGTM is too risky.

The list of potential infrastructure projects you've come up with is a good one. But I'm not suggesting they all automatically become "1-LGTM" projects. This PR's language makes that optional and case-by-case: hopefully that's clear. If the standard 2-LGTM policy is workable for a given infrastructure project, it should be used. Otherwise it could be exempted, which probably requires maintaining a list of such projects.

Again, I don't know that this applies to much in practice beyond jenkins-jobs, and my focus is on keeping the rules simple and comprehensible. If jenkins-jobs is the only one crying out for this kind of treatment currently, perhaps we start with that. (Although is it really broken today, and do we need to fix it?)

And instead of, or perhaps in addition to this proposal, I hope we understand the need for broader expertise on some of these infrastructure projects.

arschles:
I wrote a whole novel in here, then deleted it and didn't want to butt in, but still felt like I should say _something_ about my opinion, so here goes...

I think we especially need every infrastructure-related repo to require 2 LGTMs. These repositories contain code that we need to run properly and performantly to enable every developer (including 3rd parties) to make progress. Infrastructure code by definition is code that enables all other progress, and I believe we should treat it at least as an extremely important part of our organization.

bacongobbler:
so what's the story, morning glory? shall we take it to a vote?

krancour:
I vote note exemptions to the two LGTM rule for infra projects. For prototype projects-- yes. But not for infra. (Unless it's prototype infra, I guess.)

mboersma:
I would prefer not to create this exemption as well. Others?

mboersma:
It doesn't appear there is a consensus to make this change, so I'll close it. We can revisit this if / when infrastructure projects become a bottleneck again.

PR #321, opened 2016-06-14
docs(submitting-a-pull-request.md): add info on testing paired commits

vdice:
Ref https://github.com/deis/jenkins-jobs/issues/54
Ref https://github.com/deis/jenkins-jobs/pull/104

### DIFF
--- a/src/contributing/submitting-a-pull-request.md
+++ b/src/contributing/submitting-a-pull-request.md
@@ -26,6 +26,26 @@ See [testing](testing.md) for more information.
 
 Changes to any Deis Workflow component that could affect a user's experience also require a change or addition to the relevant documentation. For most Deis components, this involves updating the component's _own_ documentation. In some cases where a component is tightly integrated into [deis/workflow][workflow], its documentation must also be updated.
 
+## Cross-repo commits
+
+If a pull request is part of a larger piece of work involving one or more additional commits in other Workflow repositories, these commits can be referenced in the last PR to be submitted.  The downstream [e2e test job](https://ci.deis.io/job/workflow-test-pr/) will then supply every referenced commit to the test runner so it can source the necessary Docker images for inclusion in the generated Workflow chart to be tested.
+
+For example, consider paired commits in [deis/controller](https://github.com/deis/controller) and [deis/workflow-e2e](https://github.com/deis/workflow-e2e).  The commit body for the first PR in `deis/workflow-e2e` would look like:
+
+```
+feat(foo_test): add e2e test for feature foo
+
+[skip e2e] test for controller PR #42
+```
+Adding `[skip e2e]` forgoes the e2e tests on this commit. This and any other required PRs aside from the final PR should be submitted first, so that their respective build and image push jobs run.
+
+Lastly, the final PR in `deis/controller` should be created with the required commits listed, for use by the downstream e2e run.
+
+```
+feat(foo): add feature foo
+
+Requires workflow-e2e#abc1234
+```
 
 ## Code Standards
 


bacongobbler:
This workflow slightly confuses me. So we need to require a workflow-e2e commit to require in the controller commit, but the workflow-e2e commit requires a controller commit? Is this a chicken-and-egg problem? :)

vdice:
@bacongobbler indeed, the workflow is still a bit confusing and there still is a chicken-and-egg problem.  Definitely open to suggestions/improvements.  Speaking of which, I mention the conundrum on the issue tracking these changes and would love feedback: https://github.com/deis/jenkins-jobs/issues/54#issuecomment-225651539

bacongobbler:
thanks! I'll try to soak that all in and come up with a response.

krancour:
fwiw, I am confused by this as well.

@vdice, I definitely have not put as much thought into this topic as you have, but off the top of my head, what about this...

What if, instead of requiring the two commits to reference one another as indicated in this doc change, only _one_ had to reference the other, and the _other_ could have something in the body of the commit indicating _not_ to test it (because the test will be run by the first)?

When a job completes for any commit that references another, the test results for _both_ can be updated.

Chances are quite high this is a garbage idea, but it's a thought.

vdice:
@krancour thank you for that idea!  Indeed, I feel like this is a good approach (in fact, I had a similar thought in https://github.com/deis/jenkins-jobs/issues/54#issuecomment-225651539).  

At first, I figured we could use the `[skip ci]` message in the commit bodies for the PRs we didn't want tests running for, but unfortunately, this bypasses the entire ci pipeline, including of course the initial build/image push job, which we need to get those images out to the registries.  However, I can add logic in the existing PR related to this [change](https://github.com/deis/jenkins-jobs/pull/104) to search for, say, `[skip e2e]` which would then kick off said build/push jobs _but_ prevent the downstream e2e job from running.  

In the end, just one PR would actually kick off the full/normal ci pipeline, once all required commits/PRs have been built and images deployed, and the results of that e2e run could indeed be shared with the rest of the aforementioned PRs.  

If this sounds like a good way forward, I would add the change in https://github.com/deis/jenkins-jobs/pull/104 and adjust the documentation here to:

>For example, consider paired commits in [deis/controller](https://github.com/deis/controller) and [deis/workflow-e2e](https://github.com/deis/workflow-e2e).  The commit body for the PR in `deis/controller` would look like:

```
feat(foo): add feature foo

Requires workflow-e2e#abc1234
```
>And the commit body for the paired commit in `deis/workflow-e2e` would look like:

```
feat(foo_test): add e2e test for feature foo

[skip e2e] test for controller PR #42
```
>Adding `'[skip e2e]'` forgoes the e2e tests on this commit.  This and any other required PRs aside from the controller PR above should be submitted first, so that their respective build and image push jobs run.  Lastly, the 'final' PR in controller should be created without the `'[skip e2e]'` and with the required commits listed, for use by the actual e2e run.

krancour:
@vdice mostly sounds good. As I said, this was an idea that just came to the top of my head. After thinking about it more, here are two things I think we'd need to work through:

1. What about PRs that contain multiple commits?
2. What happens when a PR that skips e2e tests is amended? e2e tests ("owned" by a related commit) wouldn't get kicked off anew. :(

vdice:
@krancour good questions,
>What about PRs that contain multiple commits?

I guess each commit would either need to list the same `Required` commits _or_ include the `[skip e2e]`, depending on PR 'type' as mentioned previously

>What happens when a PR that skips e2e tests is amended? e2e tests ("owned" by a related commit) wouldn't get kicked off anew. :(

You're right... either way, there's some manual intervention needed.  I see two options:

  1. Whilst amending the PR commit that previously skipped e2e, one could remove the `[skip e2e]` and add all of the required commits to actually kick off the full pipeline, or,
  2. Also amend the PR that _doesn't_ skip e2e with the latest commit sha from the amended PR -- when that is pushed, it will kick off the pipeline (and hence e2e) again w/ the updated commit(s).

krancour:
> Also amend the PR that doesn't skip e2e with the latest commit sha from the amended PR

That's another dimension to this I'd overlooked. You are right... if the one that skips e2e is amended, its sha changes and the one that references it ("owns" the e2e test run) would thus have to be amended as well to reference that new sha.

In the end, I think that works. My only remaining fear would be this manual action that is required after a PR is amended being too easily forgotten.

vdice:
ping @krancour and @bacongobbler -- this PR has been updated per the conversation above (and correlating change in https://github.com/deis/jenkins-jobs/pull/104).  Please revisit when convenient.  Thanks!

PR #322, opened 2016-06-14
docs(applications): add documentation for healthchecks

bacongobbler:
This is a summarized write-up of the proposal made in deis/controller#603. Note that I'm still a bit iffy on how the client should handle the [HTTP header section](http://kubernetes.io/docs/user-guide/liveness/) in the user guide, but it's up for pondering!

### DIFF
--- a/src/applications/managing-app-configuration.md
+++ b/src/applications/managing-app-configuration.md
@@ -54,68 +54,123 @@ Detachments can be performed with `deis config:unset`.
 
 ## Custom Health Checks
 
-By default, Workflow only checks that your application containers start in
-their Pod. If you would like Kubernetes to respond to appliation health, you
-may add a health check by configuring URL, initial delay, and timeout
-values.
+By default, Workflow only checks that the application starts in their Container. If it is preferred
+to have Kubernetes respond to application health, a health check may be added by configuring a
+health check probe for the application.
 
-The health checks are implemented as [Kubernetes container probes][kubernetes-probes]. Currently only HTTP GET is supported.
+The health checks are implemented as [Kubernetes container probes][kubernetes-probes]. A `liveness`
+and a `readiness` probe can be configured, and each probe can be of type `httpGet`, `exec`, or
+`tcpSocket` depending on the type of probe the container requires.
 
-Available configuration options for health checks are the following, including defaults:
+A liveness probe is useful for applications running for long periods of time, eventually
+transitioning to broken states and cannot recover except by restarting them.
 
-**HEALTHCHECK_URL**
+Other times, a readiness probe is useful when the container is only temporarily unable to serve,
+and will recover on its own. In this case, if a container fails its readiness probe, the container
+will not be shut down, but rather the container will stop receiving incoming requests.
 
-Path in the application to use for health check, such as /healthz - Query Parameters are not supported.
+`httpGet` probes are just as it sounds: it performs a HTTP GET operation on the Container. A
+response code inside the 200-399 range is considered a pass.
 
-This value needs to be set for any health check to happen. Needs to accept a HTTP GET request and return a HTTP status between 200-399.
+`exec` probes run a command inside the Container to determine its health, such as
+`cat /var/run/myapp.pid` or a script that determines when the application is ready. An exit code of
+zero is considered a pass, while a non-zero status code is considered a fail.
 
-**HEALTHCHECK_TIMEOUT**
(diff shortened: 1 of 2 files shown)


src/applications/managing-app-configuration.md
```diff
@@ -54,68 +54,106 @@ Detachments can be performed with `deis config:unset`.
...
-
-**HEALTHCHECK\_INITIAL_DELAY**
-
-Number of seconds before health checks starts checking the application after the start of the pod.
-
-This is useful to set if the application takes a while to get setup, due to migrations, cache warming or otherwise. Prevents Kubernetes (and Deis Workflow) from replacing an application pod in its setup phase.
-
-Defaults to 50 seconds
+```
+$ deis healthchecks:set liveness httpGet / 80 --initial-delay-timeout 30 --timeout-seconds 1
```
kmala: i think path should be optional same as kubernetes keeps it and defaults to /

bacongobbler: so like `deis healthchecks:set liveness httpGet 80`, but also allow an optional argument to specify a path via `deis healthchecks:set liveness httpGet 80 /index.html`?

kmala: like `deis healthchecks:set liveness httpGet 80 --path /index.html` how we are treating other optional parameters

bacongobbler: makes sense. I'll make that change.

src/applications/managing-app-configuration.md
```diff
@@ -54,68 +54,122 @@ Detachments can be performed with `deis config:unset`.
...
+`tcpSocket` depending on the type of probe the container requires.
 
-Available configuration options for health checks are the following, including defaults:
+A liveness probe is useful for applications running for long periods of time, eventually
+transitioning to broken states and cannot recover except by restarting them.
 
-**HEALTHCHECK_URL**
+Other times, a readiness probe is useful when the container is only temporarily unable to serve,
+and will recover on its own. In this case, if a container fails its readiness probe, the container
+will not be shut down, but rather the container will stop receiving incoming requests.
```
helgi: Would be good to also make the distinction that `readiness` is used more so during `boot time` and `liveness` is a day to day checks (I'm wording this poorly but I hope you get the gist)

src/applications/managing-app-configuration.md
```diff
@@ -54,68 +54,122 @@ Detachments can be performed with `deis config:unset`.
...
 
-Path in the application to use for health check, such as /healthz - Query Parameters are not supported.
+`httpGet` probes are just as it sounds: it performs a HTTP GET operation on the Container. A
+response code inside the 200-399 range is considered a pass.
 
-This value needs to be set for any health check to happen. Needs to accept a HTTP GET request and return a HTTP status between 200-399.
+`exec` probes run a command inside the Container to determine its health. An exit code of zero is
+considered a pass, while a non-zero status code is considered a fail.
```
helgi: May want to include that the user can run a command such as `env` or to run a script they may have in their application. I know running a command implies both but I want to be very clear what is possible

helgi:
Minor suggestions but otherwise LGTM

src/applications/managing-app-configuration.md
bacongobbler: I got what you're saying. I ripped this straight off the kubernetes docs but I can add more examples on what is more applicable in certain situations

src/applications/managing-app-configuration.md
bacongobbler: How would you suggest we expand on this? Can't think of anything off the top of my head on which we could make "run a command inside the container" more clearer than it already is :)

helgi: > `exec` probes run a command inside the Container to determine its health, such as running `ps aux | grep <my process>` or run a script within your application that determines when the application is ready

Or something like that

kmala:
@bacongobbler I think it would be better to add some documentation about the current support for the previous configuration using the config:set

bacongobbler:
@kmala I'd prefer to not document legacy code. Existing users can still use it, but we shouldn't point out *how* to use it if we're trying to get away from that.

src/applications/managing-app-configuration.md
bacongobbler: added "A liveness probe is useful for applications running for long periods of time". That should make it more clear.

PR #329, opened 2016-06-17
docs(release_checklist.md): return to using immutable tags

vdice:
One of the outcomes from the last `v2.0.0` release (and subsequent retrospective) was
to return to the use of immutable tags in a release chart as opposed to the
mutable/'pretty' `$WORKFLOW_RELEASE` tags.

This change therefore represents the associated docs update.


arschles:
LGTM3

PR #357, opened 2016-06-29
feat(api): add v2.1 api documentation.

Joshua-Anderson:
@helgi Did I overlook any API changes for this release (see the `What's New` section)
I just skimmed at the [commit diff between v2.0.0 and master](https://github.com/deis/controller/compare/v2.0.0...master?expand=1).

### DIFF
--- a/src/reference-guide/controller-api/v2.1.md
+++ b/src/reference-guide/controller-api/v2.1.md
@@ -1,0 +1,1409 @@
+# Controller API v2.1
+
+This is the v2.1 REST API for the Controller.
+
+## What's New
+
+**New!** `healthcheck` field in configuration, deprecates the `HEALTHCHECK_*` environment variables.
+
+**New!** Unsetting a configuration variable that does not exist will return a 422.
+
+**New!** Creating an identical sequential release returns a 409 rather than create a no-op release.
+
+
+## Authentication
+
+### Register a New User
+
+Example Request:
+
+```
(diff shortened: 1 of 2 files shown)


helgi:
`healthcheck` is a new field within the `Config` endpoint, i see you added that in the examples but not what's new but I suppose we never really added any granularity if we have nested structures

Joshua-Anderson:
I added a message about the `healthchecks` field and removed `X-Deis-Release` headers that were entirely removed from workflow.

src/reference-guide/controller-api/v2.1.md
```diff
@@ -0,0 +1,1409 @@
+# Controller API v2.1
+
+This is the v2.1 REST API for the Controller.
+
+## What's New
+
+**New!** `healthchecks` field in configuration, deprecates the `HEALTHCHECK_*` environmental variables.
```
helgi: It needs to be singular

bacongobbler: s/environmental/environment :)

Joshua-Anderson:
Fixed comments on healthchecks.

krancour:
@Joshua-Anderson shouldn't there also be a change in `mkdocs.yml` to add this to the doc tree?

bacongobbler:
I'm surprised that mkdocs didn't catch that, but yes you are correct @krancour.

Joshua-Anderson:
@krancour Fixed!

PR #376, opened 2016-07-18
doc(dockerfiles): Add Bash to requirements list of Dockerfile/image

klpx:
closes https://github.com/deis/deis/issues/5064


src/applications/using-docker-images.md
```diff
@@ -26,6 +26,7 @@ In order to deploy Docker images, they must conform to the following requirement
 * The Dockerfile must use the `EXPOSE` directive to expose exactly one port.
 * That port must be listening for an HTTP connection.
 * The Dockerfile must use the `CMD` directive to define the default process that will run within the container.
+* If Procfile is used then Docker image should contain Bash for successfull run.
```
bacongobbler: Could you please re-word this to

```
The Docker image must contain [bash](https://www.gnu.org/software/bash/) to run processes
```

Ditto for the other line as well. Thanks!

klpx: Is it need to do the same in deis/deis?

PR #377, opened 2016-07-18
docs(install-workflow): note that k8s 1.2.x is required

mboersma:
Until https://github.com/kubernetes/kubernetes/issues/28616 and https://github.com/kubernetes/kubernetes/issues/28750 are fixed, Deis Workflow cannot start successfully on Kubernetes 1.3.x.

### DIFF
--- a/src/installing-workflow/system-requirements.md
+++ b/src/installing-workflow/system-requirements.md
@@ -4,7 +4,7 @@ To run Deis Workflow on a Kubernetes cluster, there are a few requirements to ke
 
 ## Kubernetes Versions
 
-Deis Workflow has been tested with the **Kubernetes v1.2** release line. It is incompatible with Kubernetes v1.1 and earlier.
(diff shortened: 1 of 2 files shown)


src/installing-workflow/system-requirements.md
```diff
@@ -4,7 +4,7 @@ To run Deis Workflow on a Kubernetes cluster, there are a few requirements to ke
 
 ## Kubernetes Versions
 
-Deis Workflow has been tested with the **Kubernetes v1.2** release line. It is incompatible with Kubernetes v1.1 and earlier.
+Deis Workflow has been tested with the **Kubernetes v1.2** release line. It is incompatible with Kubernetes v1.1 and earlier. Kubernetes v1.3.0 introduces [a bug mounting secrets]() which prevents Deis Workflow from starting; hopefully this will be fixed in future releases of Kubernetes.
```
vdice: `s/a bug mounting secrets/a bug when mounting` secrets so that the link will work, right?

PR #378, opened 2016-07-18
docs(base.html): remove link to Deis 1.0 docs

mboersma:
I think the distracting big red banner about "You're viewing the latest release...for 1.0 click here" can be scrapped now that 1.0 has been in maintenance mode for a while. But maybe I'm jumping the gun? ping @slack.


flynnduism:
I think enough time has passed - LGTM1

sgoings:
The flow into Deis v2 docs from Deis.com makes the destination (v1.0 vs v2.0) pretty clear to me.

Issue #372, opened 2016-07-13
Pods fail to mount secret on k8s 1.3.0 + GKE 


mboersma:
Builder, database, minio, and registry all mount the "objectstorage-keyfile" secret volume. In k8s 1.3 on GKE, this began to fail (see the final Events listing):
```
$ kubectl --namespace=deis describe po deis-database-dyqyu 
Name:		deis-database-dyqyu
Namespace:	deis
Node:		gke-mboersma-default-pool-21df92ab-lh7d/10.240.0.16
Start Time:	Wed, 13 Jul 2016 15:28:28 -0600
Labels:		app=deis-database
Status:		Pending
IP:		
Controllers:	ReplicationController/deis-database
Containers:
  deis-database:
    Container ID:	
    Image:		quay.io/deisci/postgres:canary
    Image ID:		
    Port:		5432/TCP
    QoS Tier:
      memory:		BestEffort
      cpu:		BestEffort
    State:		Waiting
      Reason:		ContainerCreating
    Ready:		False
    Restart Count:	0
    Readiness:		exec [is_running] delay=30s timeout=1s period=10s #success=1 #failure=3
    Environment Variables:
      DATABASE_STORAGE:	minio
Conditions:
  Type		Status
  Initialized 	True 
  Ready 	False 
  PodScheduled 	True 
Volumes:
  database-creds:
    Type:	Secret (a volume populated by a Secret)
    SecretName:	database-creds
  objectstore-creds:
    Type:	Secret (a volume populated by a Secret)
    SecretName:	objectstorage-keyfile
  deis-database-token-ovk2r:
    Type:	Secret (a volume populated by a Secret)
    SecretName:	deis-database-token-ovk2r
Events:
  FirstSeen	LastSeen	Count	From							SubobjectPath	Type		Reason		Message
  ---------	--------	-----	----							-------------	--------	------		-------
  4m		4m		1	{default-scheduler }							Normal		Scheduled	Successfully assigned deis-database-dyqyu to gke-mboersma-default-pool-21df92ab-lh7d
  2m		2m		1	{kubelet gke-mboersma-default-pool-21df92ab-lh7d}			Warning		FailedMount	Unable to mount volumes for pod "deis-database-dyqyu_deis(bd127bc0-4940-11e6-af68-42010af001d0)": timeout expired waiting for volumes to attach/mount for pod "deis-database-dyqyu"/"deis". list of unattached/unmounted volumes=[objectstore-creds]
[...]

PR #386, opened 2016-07-20
Add instructions to downgrade GKE node versions

foxycoder:
After spending too much time on installing Workflow, @kmala pointed out in Slack that I was probably running on k8s 1.3.x. Which is the default. I checked in the Console and there is no other way than to create the new cluster then downgrade it.

Hope with this (temporary) fix in the docs, other people will not make the same mistake. 😉 

See a preview of my additions below:

## Check Kubernetes version

[](http://cd.sseu.re/2016_07_20_16_25_45_afsqg.jpg)

After the cluster is created, check the node version. If you are on 1.3.x, **downgrade the node version to 1.2.5**. See [Kubernetes Versions](https://deis.com/docs/workflow/installing-workflow/system-requirements/#kubernetes-versions) under System Requirements for more details.


mboersma:
For reference, here is the bug that makes this downgrade necessary for now: https://github.com/deis/workflow/issues/372

slack:
Thanks for the clarification @foxycoder would you mind if we mirrored the screenshot under deis.com?

smothiki:
@mboersma we should merge this and rebase the https://github.com/deis/workflow/pull/387

mboersma:
> we should merge this

The image URL needs to be updated though, so deis.com isn't relying on an unrelated web host. We could merge this and then quickly follow up by re-hosting the image in a different PR. Whatcha think @slack?

foxycoder:
Hi sorry I moved the image the repo. Hope this helps. :)

slack:
Great, thanks!

PR #395, opened 2016-07-21
doc(deploying-apps.md): add note about known issue in controller

arschles:
when deployments are turned on, the controller can crash on startup in
a very specific case

Ref https://github.com/deis/controller/issues/903

### DIFF
--- a/src/applications/deploying-apps.md
+++ b/src/applications/deploying-apps.md
@@ -103,9 +103,10 @@ each having multiple ReplicaSets (one per release) which in turn manage the Pods
 !!! Known Issue
 		If `DEIS_KUBERNETES_DEPLOYMENTS` is enabled and there is a deployment in progress when the
 		controller starts, the controller will crash until the deployment is finished. This issue can
-		specifically manifest itself if the controller crashes while you are scaling an app. The scale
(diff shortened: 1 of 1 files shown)


src/applications/deploying-apps.md
```diff
@@ -100,6 +100,14 @@ Where you will see differences while using the CLI is `deis ps:list` will output
 Behind the scenes your application deploy will be built up of a Deployment object per process type,
 each having multiple ReplicaSets (one per release) which in turn manage the Pods running your application.
 
+!!! Known Issue
+		If `DEIS_KUBERNETES_DEPLOYMENTS` is enabled and there is a deployment in progress when the
```
bacongobbler: Needs an extra newline to render this properly :)

arschles: Do you mean an extra newline below `!!! Known Issue`? Otherwise, I want the entire text to be inside the warning box

bacongobbler: Yes, that's what I meant :)

bacongobbler: Ah derp I'm misleading you. Never mind. Carry on!

PR #405, opened 2016-07-25
doc(applications): add pre-compile/post-compile hook docs

bacongobbler:
refs https://github.com/deis/slugbuilder/pull/86

### DIFF
--- a/src/applications/using-buildpacks.md
+++ b/src/applications/using-buildpacks.md
@@ -99,6 +99,21 @@ To use a custom buildpack, set the `BUILDPACK_URL` environment variable.
 
 On your next `git push`, the custom buildpack will be used.
 
+
+## Compile Hooks
+
+Sometimes, an application needs a way to stop or check if a service is running before building an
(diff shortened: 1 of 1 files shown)


src/applications/using-buildpacks.md
```diff
@@ -99,6 +99,21 @@ To use a custom buildpack, set the `BUILDPACK_URL` environment variable.
 
 On your next `git push`, the custom buildpack will be used.
 
+
+## Compile Hooks
+
+Sometimes, an application needs a way to stop or check if a service is running before building an
+app, and sometimes it requires notifying a service that the [Builder][] has finished compiling the
```
mboersma: This repeats "sometimes" twice. Maybe rephrase it like:
> Sometimes, an application needs a way to stop or check if a service is running before building an app, which may require notifying a service that...

bacongobbler:
Sorry, missed that comment @mboersma. I'll follow up. :)

PR #410, opened 2016-07-26
docs(roadmap): rewrite for new release process

mboersma:


Joshua-Anderson:
It looks like `deisrel changelog` is still in use. Are there any other parts of `deisrel` still being used?

mboersma:
> Are there any other parts of deisrel still being used?

I think the new [check component versions](https://github.com/deis/deisrel/pull/122) `deisrel` functionality will be leveraged here, but I don't know otherwise. I'm still catching up with changes and trying to reconcile our [release goals](https://gist.github.com/jackfrancis/408006654fa9adac4aa48e4a4ed42a36) with reality. I hope to have these docs completed by tomorrow a.m.

src/roadmap/releases.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+### Step 4: Publish Release to Versions Server
+
+Finally, use the Workflow Manager API publishing tool to add the new component release to the
+versions server. This step is described in the workflow-manager-api-publish docs and must be done
+by a Deis maintainer.
+
+
+## How to Release Workflow
```
vdice: Wondering if we should add info as to using `deisrel` to see the `clean/dirty` output of all the components, for informing what tags are supplied in the release charts's `generate_params.toml` in Step 3. below?  See https://github.com/deis/deisrel/pull/122#issuecomment-235944997 for background...

mboersma: We definitely should, but that code hadn't landed yet. I'll update the PR in a few.

bacongobbler:
Awesome. Very clear and concise docs, as always. :+1:

PR #438, opened 2016-08-09
docs(applications): add deis routing docs

bacongobbler:
refs https://github.com/deis/controller/pull/934

### DIFF
--- a/src/applications/domains-and-routing.md
+++ b/src/applications/domains-and-routing.md
@@ -1,11 +1,11 @@
 ## Domains and Routing
 
-You can use `deis domains` to add or remove custom domains to your application:
+You can use `deis domains` to add or remove custom domains to the application:
 
     $ deis domains:add hello.bacongobbler.com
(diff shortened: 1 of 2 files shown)


src/applications/domains-and-routing.md
```diff
@@ -20,3 +20,13 @@ appname to the old one:
     ("start-of-authority") records. It is highly recommended that you bind a subdomain to
     an application, however you can work around this by pointing the @ record to the
     address of the load balancer (if any).
+
+To add or remove your application from the routing mesh, use `deis routing`:
+
+    $ deis routing:disable
+    Disabling routing for finest-woodshed... done
+
+This will make your application unreachable through the [Router][].
```
slack: Since we reference "add or remove" it might make sense to include the enable example as well.

Should also note that when routing is "disabled" the application is still reachable via its Kubernetes service.

bacongobbler: noted

PR #443, opened 2016-08-11
feat(maintenance): Add docs for maintenance mode support

kmala:
Fixes https://github.com/deis/deis/issues/3722

This should be merged after corresponding code changes are merged.

- [x] https://github.com/deis/controller-sdk-go/pull/75
- [x] https://github.com/deis/controller/pull/966
- [x] https://github.com/deis/workflow-cli/pull/171
- [x] https://github.com/deis/router/pull/232

### DIFF
--- a/src/applications/managing-app-lifecycle.md
+++ b/src/applications/managing-app-lifecycle.md
@@ -98,6 +98,22 @@ From deis-controller.local:peachy-waxworks
  * [new branch]      master     -> deis/master
 ```
 
+
+## Application Maintenance
+
+Maintenance mode for applications is useful to perform certain migrations or upgrades during which we don't want to serve client requests. Deis Workflow supports maintenance mode for an app during which the access to the app is blocked. Blocking access to the application means all the requests to the app are served with an error code of `503` and a static maintenance page by the router but the app will still be running and one-off commands can still be run. Currently the maintenance page is not configurable and is present as part of the router component.
+
+To enable maintenance mode for app, use `deis maintenance`:
+
+    $ deis maintenance:on
+    Enabling maintenance for drafty-zaniness... done
+
+This will make the [router][] answer all requests to the application with a 503, although the app is still running. To disable maintenance mode:
+
+    $ deis maintenance:off
+    Disabling maintenance for drafty-zaniness... done
+
+
 ## Application Troubleshooting
 
 Applications deployed on Deis Workflow [treat logs as event streams][]. Deis Workflow aggregates `stdout` and `stderr`
@@ -124,3 +140,4 @@ Use `deis logs` to view the log output from your deployed application.
 [treat logs as event streams]: http://12factor.net/logs
 [use one-off processes for admin tasks]: http://12factor.net/admin-processes
 [Procfile]: http://ddollar.github.io/foreman/#PROCFILE
+[router]: ../understanding-workflow/components.md#router


src/applications/managing-app-lifecycle.md
```diff
@@ -98,6 +98,22 @@ From deis-controller.local:peachy-waxworks
  * [new branch]      master     -> deis/master
 ```
 
+
+## Application Maintenance
+
+Maintenance mode for applications is useful to perform certain migrations or upgrades during which we don't want to serve client requests. Deis Workflow supports maintenance mode for app during which the access to the app is blocked. Blocking access to the application means all the requests to the app are served with error code of `503` and static maintenance page by the router but the app will still be running and one-off commands can still be run on them.
+
+To enable maintenance mode for app, use `deis routing`:
```
bacongobbler: use `deis maintenance`*

smothiki:
Looks good to me

src/applications/managing-app-lifecycle.md
```diff
@@ -98,6 +98,22 @@ From deis-controller.local:peachy-waxworks
...
+## Application Maintenance
+
+Maintenance mode for applications is useful to perform certain migrations or upgrades during which we don't want to serve client requests. Deis Workflow supports maintenance mode for app during which the access to the app is blocked. Blocking access to the application means all the requests to the app are served with error code of `503` and static maintenance page by the router but the app will still be running and one-off commands can still be run on them.
+
+To enable maintenance mode for app, use `deis maintenance`:
+
+    $ deis maintenance:on
+    Enabling maintenance for drafty-zaniness... done
+
+This will make all the request to the application serve 503 through the [Router][], but the application is still running. To disable maintenance mode:
```
mboersma: Do markdown links need to match in case as well? If so, `router` should be lower-case here. Also I might rephrase it a bit:
```
This will make the [router][] answer all requests to the application with a 503, although the app
is still running. To disable maintenance mode:
```

src/applications/managing-app-lifecycle.md
```diff
@@ -98,6 +98,22 @@ From deis-controller.local:peachy-waxworks
  * [new branch]      master     -> deis/master
 ```
 
+
+## Application Maintenance
+
+Maintenance mode for applications is useful to perform certain migrations or upgrades during which we don't want to serve client requests. Deis Workflow supports maintenance mode for app during which the access to the app is blocked. Blocking access to the application means all the requests to the app are served with error code of `503` and static maintenance page by the router but the app will still be running and one-off commands can still be run on them.
```
mboersma: Need to add articles (a, an, the) before three nouns:
> maintenance mode for *an* app
> with *an* error code of `503` and *a* static maintenance page

Also I would remove the last two words "on them".

src/applications/managing-app-lifecycle.md
```diff
@@ -98,6 +98,22 @@ From deis-controller.local:peachy-waxworks
  * [new branch]      master     -> deis/master
 ```
 
+
+## Application Maintenance
+
+Maintenance mode for applications is useful to perform certain migrations or upgrades during which we don't want to serve client requests. Deis Workflow supports maintenance mode for an app during which the access to the app is blocked. Blocking access to the application means all the requests to the app are served with an error code of `503` and a static maintenance page by the router but the app will still be running and one-off commands can still be run.
```
helgi: You'll want to mention that it is not a configurable maintenance page right now

kmala: we have said it is a static maintenance page by the router and never mentioned about configuration. Does it convey that can't be configured

helgi: I feel it's ambiguous enough that a person would go looking in the router how to change it

kmala: Updated.

PR #444, opened 2016-08-11
bug(docs): clarify object storage requirements

### LINKED ISSUE #436: Clarify that storage must be configured before installing workflow
I was under the impression that I could configure the ephemeral storage for Workflow after I installed it, but was not aware it had to be set up before installing.  I felt like since it was a section after the install, it wasn't necessary at the immediate time

image

It would be helpful to add a section regarding Storage on the installation page:
https://deis.com/docs/workflow/installing-workflow/

slack:
Fixes #436


slack:
2016-08-11 at 13 04

2016-08-11 at 13 05

PR #454, opened 2016-08-16
doc(storage): cleanup storage docs

slack:
The bullet-ocolypse was getting a bit out of hand! Refactor into a table. Then
some small clean-ups for style, etc.


slack:
objstorage

PR #455, opened 2016-08-16
chore(docs): update docs to reflect fixed issue

jeroenvisser101:
Since https://github.com/deis/controller/issues/903 is fixed, this should not be an issue anymore, right?


helgi:
Yeah - Thanks. Was leaving this knocking around until the release this week but now is good too :)

jeroenvisser101:
ah, I came across it when reading, no big deal :)

PR #461, opened 2016-08-18
feat(_scripts): build every tagged version of workflow docs

### LINKED ISSUE #351: Documentation should target latest version of deis, not the development version.
Right now, every PR targets master, regardless of whether the PR addresses an unreleased version of deis or it targets the current release.

One solution that could fix this is that `master` branch targets the current release of deis and is currently in production. Then a second branch, called `future` targets the next version of deis and is merged into master when that next version is released.

There are lots of ways to solve this problem, so what does everyone think about this?

bacongobbler:
Since gutenberg uses rsync to pull in the projects for building the docs, the tags *should* be available for building.

ping @slack; Would this be the right way to do it?

closes #351

### DIFF
--- a/_scripts/gutenbuild
+++ b/_scripts/gutenbuild
@@ -6,3 +6,17 @@ set -x
 build_dest=$1
 
 PATH=$HOME/.local/bin:$PATH BUILDDIR=${build_dest} make deps build
+
+# build again, this time to /en/dev
+dev_build_dest=$1/en/dev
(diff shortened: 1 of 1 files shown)


bacongobbler:
punting

bacongobbler:
ping @slack, you mentioned you had a better idea

bacongobbler:
ping

slack:
This should be fine, alternatives aren't worth the effort!

PR #474, opened 2016-08-24
docs(ssl-certificates): add docs on deis tls:enable

bacongobbler:


src/applications/ssl-certificates.md
```diff
@@ -124,6 +124,21 @@ Pay attention to the output. It should print `SSL certificate verify ok`. If it
 like `common name: www.example.com (does not match 'www.somedomain.com')` then something is not
 configured correctly.
 
+## Enforcing SSL at the Router
+
+If it is required to have all HTTP requests be redirected to HTTPS, TLS can be enforced at the
```
Joshua-Anderson: >If it is required to have all HTTP requests be redirected to HTTPS, TLS can be enforced at the

Maybe `To require all HTTP requests be redirected to HTTPS` is a little clearer?

PR #478, opened 2016-08-25
docs(upgrade): reminder to upgrade deis client

olalonde:


Joshua-Anderson:
Deis no longer shows warnings when using an older client with an newer controller, since it should be safe.

That being said, upgrading for the new features is probably a good idea.

src/managing-workflow/upgrading-workflow.md
```diff
@@ -173,3 +173,11 @@ deis-workflow-manager-2654760652-kitf9   1/1       Running   0          5m
 [configuring object storage]: ../installing-workflow/configuring-object-storage.md
 [keeper manifests]: http://helm-classic.readthedocs.io/en/latest/awesome/#keeper-manifests
 [minio]: https://github.com/deis/minio
+
+### Step 7: Upgrade the deis client
+
+Your deis platform users should now upgrade their deis client to avoid getting `WARNING: Client and server API versions do not match. Please consider upgrading.` warnings.
+
+```
+curl -sSL http://deis.io/deis-cli/install-v2.sh | bash && mv deis $(which deis)
```
Joshua-Anderson: You also want to `chmod +x deis`.

I might not try to move the cli. I put deis in `/user/local/bin`, so moving without sudo is going to fail.

olalonde:
@Joshua-Anderson ah ok, I got the error with v2.2.0 client

src/managing-workflow/upgrading-workflow.md
olalonde: I didn't have to `chmod +x deis`, I think the install script does it.

olalonde:
@mboersma oops, should have been step #8 :( https://deis.com/docs/workflow/managing-workflow/upgrading-workflow/#step-7-upgrade-the-deis-client https://deis.com/docs/workflow/managing-workflow/upgrading-workflow/#step-7-verify-upgrade

PR #483, opened 2016-09-01
feat(autoscale): add documentation for the autoscale functionality

### LINKED ISSUE #403: [proposal] Support process type autoscaling
Developers should be able to define an autoscaling policy per application process type. Under the covers, Workflow should use `HorizontalPodAutoscaling`.

Pre-req: Kubernetes Deployments

Examples:

- `deis autoscale:set web --min=2 --max=10 --cpu-percent=80`
- `deis autoscale:unset web`
- `deis autoscale:list`

If an operator users `deis scale web=5` when the application is set to autoscale, the replica count for the deployment should be set to that value and  the HPA should be left alone.

When recovering from a backup or controller restart the HPA should be set with known values and initial replica count should default to the last user-provided value.  Commands like `deis ps` should use Kubernetes as the source of truth for pods, rather than the postgres database.

helgi:
Closes #403

### DIFF
--- a/src/applications/managing-app-processes.md
+++ b/src/applications/managing-app-processes.md
@@ -85,9 +85,9 @@ Scaling processes... but first, coffee!
 done in 0s
 === steely-mainsail Processes
 --- cmd:
-steely-mainsail-v3-cmd-nyrim up (v3)
+steely-mainsail-cmd-3291896318-nyrim up (v3)
 --- sleeper:
-steely-mainsail-v3-sleeper-oq1jr up (v3)
+steely-mainsail-sleeper-3291896318-oq1jr up (v3)
 ```
 
 !!! note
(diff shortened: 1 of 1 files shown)


src/applications/managing-app-processes.md
```diff
@@ -159,13 +159,51 @@ Scaling processes... but first, coffee!
...
+```
+$ deis autoscale:unset web
+Removing autoscale for process type web on scenic-icehouse... done
+```
+
+For autoscaling to work CPU requests have to be specified on each application Pod (can be done via `deis limits --cpu`). This allows the autoscale policies to do the [appropriate calculations][autoscale-algo] and make decisions on when to scale up and down.
+
+Scale up can only happen if there was no rescaling within the last 3 minutes. Scale down will wait for 5 minutes from the last rescaling. That information and more can be found at [HPA algorithm page][autoscale-algo].
+
+!!! note
```
bacongobbler: this should probably be much higher up

helgi: I did debate about that - I'm fine with moving it up

PR #489, opened 2016-09-06
fix(releases): deisrel no longer requires $OLD_SHA and $OLD_RELEASE

bacongobbler:


Joshua-Anderson:
Do you want to mention that they can still be overridden with:

```
OPTIONS:
   --base-tag value  old tag where changelog starts. Optional, defaults to most recent tag
   --sha value       commit that will be tagged in the new release. Optional, defaults to most recent commit on master.
```

bacongobbler:
I think we can skip those for now. Most users releasing components won't require those flags until we start cutting patches for old minor releases (which we've never done before)

PR #491, opened 2016-09-06
chore(src/roadmap/releases.md): update jenkins-jobs var name

vdice:
From `WORKFLOW_RELEASE` to `workflowChartRelease`

Ref https://github.com/deis/jenkins-jobs/pull/230


helgi:
Would be good if we made a simple `sed` command for this if we are not automating it in deisrel

vdice:
@helgi great idea; `perl` command added.

Issue #493, opened 2016-09-06
Purge generate-changelog.sh


mboersma:
`deisrel` is now the preferred tool for generating CHANGELOG-style diffs. This orphans the `_scripts/generate-changelog.sh` script in numerous repositories.

helgi:
See https://github.com/deis/workflow/issues/464

mboersma:
@helgi oops, missed that. I'll restrict this issue to the purging of the generator script then.

Issue #464, opened 2016-08-20
CHANGELOG.md is present in most (all?) repos but hasn't been updated since 2.0


helgi:
We should either remove all CHANGELOG.md files (or use it to point people to tags) or find a way to update it along with the git tags.

I know there was some back and forth on this before for the scenario of 2.4 is released, `master` is now 2.5, however 2.4.1 needs to be released and a fix is committed to `master` and then pulled into 2.4.1 via cherry-pick - Technically the item would show up under 2.4.1 and 2.5 unless we do some more intelligent way of bringing 2.4.1 back into `master` via Changelogs, etc etc

bacongobbler:
TODO:

 - [ ] remove CHANGELOG.md
 - [ ] remove all scripts under _scripts that are changelog related (e.g. `deis-contributions-since`, `generate-changelog.sh`)
 - [ ] upload changelogs for any old component releases on github

bacongobbler:
OK, so here's an exhaustive list of PRs that once merged, check points 1 and 2:

deis/controller#1091
deis/workflow-cli#253
deis/charts#361
deis/router#264
deis/monitor#145
deis/slugrunner#55
deis/slugbuilder#114
deis/postgres#148
deis/fluentd#55
deis/registry#68
deis/workflow-e2e#327
deis/workflow-manager#102
deis/minio#118
deis/logger#119
deis/dockerbuilder#99
deis/builder#435

Optionally, deis/pkg#26 removes some cruft left over as well, but isn't an official product... More of a byproduct.

bacongobbler:
closing as all CHANGELOGs have been removed. Tracking the upload of old changelogs for previous releases in #531.

PR #513, opened 2016-09-23
chore(_scripts): remove old release tools and CHANGELOG

bacongobbler:
Now that we publish changelogs to github, the need for a CHANGELOG.md
or the tools to generate it is no longer required.

refs #464 
refs #493


helgi:
Would it not be better to have `deisrel` updating the changelog as well as creating information in tags? And if not that then point people there via CHANGELOG.md

bacongobbler:
We opted out of updating CHANGELOG due to some discussion over how we merge CHANGELOG changes using deisrel back into master. There was a chicken and egg problem so we opted out of it, which is why #464 was created.

PR #519, opened 2016-09-27
fix(gutenbuild): do not build any pre-release tags

bacongobbler:
The pre-release tags do not have the `make deps` or `make build` targets. This PR removes them from the build process.


bacongobbler:
escalating to p0 because the docs are currently breaking due to my recent change in #461

PR #522, opened 2016-09-27
docs(managing-workflow): Document PROXY protocol configuration

felixbuenemann:
This adds a section on configuring PROXY protocol with a specific example targeting K8s on AWS using ELB.

Because of the fix in deis/router#263 this PR should likely be held back until Workflow v2.6.0 is released.


felixbuenemann:
@kmala Should we mention that K8s 1.3.x or later is needed to support the `service.beta.kubernetes.io/aws-load-balancer-proxy-protocol`annotation?

kmala:
@felixbuenemann yes..it is better to mention that and also say that it can be configured manually through  http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/enable-proxy-protocol.html

felixbuenemann:
I've added an "important" info box with the K8s version requirement and rebased on master.

PR #530, opened 2016-09-30
docs(relases): simplify Workflow release step 4

mboersma:
A script in the charts repository removes much of the manual work here.

Requires deis/charts#363.

### DIFF
--- a/src/roadmap/releases.md
+++ b/src/roadmap/releases.md
@@ -164,9 +164,7 @@ For a major or minor release, copy and modify the current development charts:
 
 ```bash
 git checkout -b release-$WORKFLOW_RELEASE master
-cp -r workflow-dev workflow-$WORKFLOW_RELEASE
-cp -r workflow-dev-e2e workflow-$WORKFLOW_RELEASE-e2e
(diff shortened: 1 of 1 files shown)


src/roadmap/releases.md
```diff
@@ -201,21 +199,6 @@ deisrel $HOME/.helmc/workspace/charts/workflow-$WORKFLOW_PREV_RELEASE/tpl/genera
 Change the `generate_params.toml` file in **each** new chart as follows:
 
   1. Set all `dockerTag` values to latest releases for each component, as determined above
```
helgi: This couldn't be scripted as well?

mboersma: Not really, or at least not reliably. We don't always release the latest tagged versions of every component--for example, we pumped the brakes on fluentd changes in Workflow v2.5.0, and registry-proxy has a commit in master that shouldn't be included until there are further releasable changes.

At least for now, I think this step requires human quality control, but if we can get the process to actually match "*always* release the latest tagged components," then this could also be scripted.

PR #535, opened 2016-10-04
docs(releases): update component release steps

mboersma:
Manual tasks are now mostly gone in favor of running the `deisrel release` command.

See also deis/deisrel#140.


bacongobbler:
:shipit:

mboersma:
I updated the docs to match `deisrel release` and removed the LGTMs based on the previous script.

PR #561, opened 2016-10-18
(docs)Add instructions to preserve platform SSL credentials

seanknox:
Added upgrade instructions to preserve existing the platform SSL certificate and private key. Installations that enforce SSL for all HTTP requests via `router.deis.io/nginx.ssl.enforce` *must* do this or operators may find themselves in quite the [head-scratching moment as applications continue to function yet the CLI, grafana, and other components requiring SSL fail](https://deis.slack.com/archives/community/p1476740206006301).

### DIFF
--- a/src/managing-workflow/upgrading-workflow.md
+++ b/src/managing-workflow/upgrading-workflow.md
@@ -75,7 +75,7 @@ Fetching the new chart copies the chart from the chart cache into the helmc work
 $ helmc fetch deis/workflow-${DESIRED_WORKFLOW_RELEASE}
 ```
 
-### Step 3: Fetch credentials
+### Step 3: Fetch builder, database, and platform SSL credentials
 
 The first time Workflow is installed, Helm automatically generates secrets for the builder and database components.
 When upgrading, take care to use credentials from the running Workflow installation. The following commands export the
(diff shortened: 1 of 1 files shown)


src/managing-workflow/upgrading-workflow.md
```diff
@@ -89,6 +89,14 @@ $ kubectl --namespace=deis get secret database-creds -o yaml > ~/active-deis-dat
 $ kubectl --namespace=deis get secret builder-ssh-private-keys -o yaml > ~/active-deis-builder-secret-ssh-private-keys.yaml
 ```
 
+#### Optional: Fetch platform SSL credentials
+Installations that have installed a [platform SSL](https://deis.com/docs/workflow/managing-workflow/platform-ssl/) certificate and private key on the router will need to save those secrets locally as well. They will be installed at a later step with the other credentials saved to the local workstation.
+
+``` 
+# fetch the platform SSL certificate and private key and store on the local workstation
+$ kubectl --namespace=deis get secret deis-router-platform-cert -o yaml > ~/active-deis-router-platform-cert
```
ultimateboy: Can you please add the `.yaml` extension to this file? Down below when you go to move this into the helmc workspace, you specify a .yaml extension.

seanknox: Fixed, thanks for catching. Also amended the commit message to use the correct (I hope) format.

ultimateboy:
Thanks, @seanknox!

seanknox:
@felixbuenemann You're correct about upgrades preserving the deis namespace, router, and certificate. My use case was creating a new cluster using the same customized `workflow-v{$VERSION}` charts that annotated the router with `router.deis.io/nginx.ssl.enforce`.

felixbuenemann:
That makes sense and some people might want to create a clean cluster on each upgrade for easier rollbacks if something goes wrong.

PR #572, opened 2016-10-24
Fix minor typo

merwan:


mboersma:
For future reference, please format commit messages as described here: https://deis.com/docs/workflow/contributing/submitting-a-pull-request/
That way our CHANGELOG scripting will include it.

PR #586, opened 2016-10-27
docs(releases): workflow releases happen monthly

mboersma:
Workflow has grown into a mature and stable product, and the team at Deis thinks doing a minor release every two weeks may not be productive. This proposes to change the Deis Workflow release policy to "first Thursday of each month," but the actual change is dependent on your feedback. So please comment here as to whether or not that's a useful release policy for you!

See also deis/seed-repo#48.

### DIFF
--- a/src/roadmap/releases.md
+++ b/src/roadmap/releases.md
@@ -23,11 +23,12 @@ tools, to Docker base images, and to other Deis projects that do [semantic versi
 See "[How to Release a Component](#how-to-release-a-component)" for more detail.
 
 
-## Workflow Releases Every Two Weeks
+## Workflow Releases Each Month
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Spelling/grammar LGTM.

felixbuenemann:
I think monthly minor releases are fine.

There can always be patch releases if there are important issues that need to be fixed quickly and users that are eager to try out new features can switch to a newer ci image.

rimusz:
I agree with @felixbuenemann if there is some major patch release, yes then we should an extra monthly release.

mboersma:
>  if there is some major patch release, yes then we should an extra monthly release.

Agreed, and I kept that distinct from the monthly feature-oriented releases:
> Patch releases are created at any time, as needed.

PR #592, opened 2016-10-31
docs(index): add note on contributing

bacongobbler:
This adds a note to the homepage of the workflow documentation on how users can start contributing to the project.


bacongobbler:
Thanks! That was the intention, I just derped.

bacongobbler:
Done

PR #594, opened 2016-11-01
docs(src/roadmap/releases.md): update master changelog step

vdice:
Ref https://github.com/deis/deisrel/pull/146

### DIFF
--- a/src/roadmap/releases.md
+++ b/src/roadmap/releases.md
@@ -235,10 +235,34 @@ release is predetermined as the value of `$WORKFLOW_RELEASE`.
 
 ### Step 7: Assemble Master Changelog
 
-Each component already updated its release notes on GitHub with CHANGELOG content. The
-bodies of each component's release notes should be concatenated into a single gist. Note that there
-may be more than one release per component--and more than one set of release notes--included in the
-Workflow release.
+Each component already updated its release notes on GitHub with CHANGELOG content. We'll now
+generate the master changelog for the Workflow chart, consisting of all aforementioned component changes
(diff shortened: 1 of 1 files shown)


src/roadmap/releases.md
```diff
@@ -235,10 +235,31 @@ release is predetermined as the value of `$WORKFLOW_RELEASE`.
...
+invoking `deisrel changelog global` to get all component changes between the tag existing in the `WORKFLOW_PREV_RELEASE`
+chart and the _most recent_ release tag existing in GitHub. (Therefore, if there are any unreleased commits in
+a component repo, they will not appear here):
+
+```bash
+deisrel changelog global $HOME/.helmc/workspace/charts/workflow-$WORKFLOW_PREV_RELEASE/tpl/generate_params.toml \
+  components.json > $WORKFLOW_RELEASE
+```
+
+To get non-component repo changelogs (presumably tagged in Step 3 above), one can issue a command like the following
```
mboersma: If you do it this way, the potential is there to pull in commits that were merged *after* the tag we care about. Whereas scraping them from the actual GH release notes would always be the right content. In most cases they will agree since we do this release dance in a short period of time, but getting them from the release itself is going to be safer.

vdice: @mboersma good call -- updated with an example curling the GH api -- let me know what you think.

mboersma: That bit of shell scripting works great, I tested it locally. :+1:

bacongobbler:
Haven't tested myself but code LGTM.

Issue #570, opened 2016-10-24
Have one place to install workflow and refer it instead of duplicating


kmala:

bacongobbler:
@mboersma @vdice or @kmala do you remember why we opened this ticket (i.e. where in the docs is the duplicate information)?

kmala:
https://github.com/deis/workflow/blob/master/src/quickstart/provider/aws/install-aws.md 
https://github.com/deis/workflow/blob/master/src/quickstart/provider/gke/install-gke.md
https://github.com/deis/workflow/blob/master/src/quickstart/provider/vagrant/install-vagrant.md
https://github.com/deis/workflow/blob/master/src/quickstart/provider/azure-acs/install-azure-acs.md
All these say almost the same info and steps.

mboersma:
I *thought* the idea was to reduce the amount of copy-and-pasted content between the several "Quick Start" installation paths. `sphinx` had an `include` directive, but `mkdocs` doesn't, so that would either require some trickery or rearrangement of content into general and specific installation paths with links to each other (which I think we should avoid--it wasn't user-friendly in Deis v1 when we had that).

mboersma:
Maybe there is a Markdown pre-proceesor or `mkdocs` extension that allows for include-type blocks? In Deis v1 we had directives like this throughout the install docs to keep things [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself):

```.. include:: ../_includes/_generate-discovery-url.rst```

If there's no such mechanism, I think we should close this issue and live with the duplication. Users *really* did not appreciate being told "go here for general DNS setup instructions, then return to this AWS-specific guide," and that seems like the only other solution.

vdice:
@mboersma thanks for researching; I'll take a look as well.  

I definitely agree if there isn't such an `include` solution we should stick w/ the duplication.

bacongobbler:
https://pypi.python.org/pypi/markdown-include exists and can be installed as an extension of mkdocs. I can tackle this.

PR #599, opened 2016-11-08
docs(*): add doc on signed helm charts

vdice:
As well as self-hosted public key used to sign charts.

cc @slack @kmala


src/installing-workflow/index.md
```diff
@@ -83,6 +83,8 @@ $ helmc install workflow-v2.8.0               # injects resources into
 
     	$ helm install deis/workflow --version=v2.8.0 --namespace=deis -f <optional values file>  # injects resources into your cluster
 
+  See also our section on [Workflow chart provenance](workflow-helm-charts.md#chart-provenance)
+
```
kmala: should we add it to our other install sections in providers?

vdice: @kmala good point -- actually, hoping a follow-up PR to address https://github.com/deis/workflow/issues/570 will remove/condense all the other (duplicated) sections, so perhaps moot by then.

vdice: @kmala updated

vdice:
added the `needs manual testing` label as I'd like to see another reviewer run some of the key fetching/chart verification commands as presented in this doc PR

mboersma:
I tested as described here with `--verify`. 👍

PR #600, opened 2016-11-09
feat(charts): Add support to add annotations for router during install

kmala:
ref https://github.com/deis/router/pull/288

### DIFF
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -165,6 +165,15 @@ registry-token-refresher:
 
 router:
   dhparam: ""
+  # Any custom router annotations(https://github.com/deis/router#annotations)
+  # which need to be applied can be specified as key-value pairs under "deployment_annotations"
+  deployment_annotations:
+    #<example-key>: <example-value>
+
+  # Any custom annotations for k8s services like http://kubernetes.io/docs/user-guide/services/#ssl-support-on-aws
+  # which need to be applied can be specified as key-value pairs under "service_annotations"
+  service_annotations:
+    #<example-key>: <example-value>
 
 workflow-manager:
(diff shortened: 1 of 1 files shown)


charts/workflow/values.yaml
```diff
@@ -165,6 +165,15 @@ registry-token-refresher:
 
 router:
   dhparam: ""
+  # Any custom router annotations(https://github.com/deis/router#annotations)
+  # which needs to applied can be specified as key values pair under this deployment_annotations key
```
mboersma: I would rewrite this slightly:
`which need to be applied can be specified as key-value pairs under "deployment_annotations"`

charts/workflow/values.yaml
```diff
@@ -165,6 +165,15 @@ registry-token-refresher:
...
+  deployment_annotations:
+    #<example-key>: <example-value>
+
+  # Any custom annotations for k8s services like http://kubernetes.io/docs/user-guide/services/#ssl-support-on-aws
+  # which needs to applied can be specified as key values pair under this service_annotations key
```
mboersma: Same as above:
`which need to be applied can be specified as key-value pairs under "service_annotations"`

charts/workflow/values.yaml
```diff
@@ -165,6 +165,15 @@ registry-token-refresher:
 
 router:
   dhparam: ""
+  # Any custom router annotations(https://github.com/deis/router#annotations)
+  # which needs to be applied can be specified as key-value pairs under "deployment_annotations"
```
mboersma: `needs` should be `need` here and below. Otherwise this looks good to me.

vdice:
Other than https://github.com/deis/workflow/pull/600#discussion_r87674794, LGTM

PR #611, opened 2016-11-21
docs(*): migrate from Helm Classic to Helm

bacongobbler:
This changes all documentation to use Helm as the default tool for installing Deis
Workflow. Users of Helm Classic are urged to use https://github.com/deis/workflow-migration
to migrate from Helm Classic to Helm.

Note that I kept roadmap/releases.md intentionally left unchanged for now as we are still pumping out Helm Classic charts for v2.9.0 as far as I'm aware.

### DIFF
--- a/src/contributing/overview.md
+++ b/src/contributing/overview.md
@@ -6,19 +6,7 @@ Interested in contributing to a Deis project?  There are lots of ways to help.
 
 Find a bug? Want to see a new feature? Have a request for the maintainers? Open a Github issue in the applicable repository and we’ll get the conversation started.
 
-Our official support channels are:
-
-- GitHub issue queues:
-    - [builder](https://github.com/deis/builder/issues)
-    - [chart](https://github.com/deis/charts/issues)
-    - [database](https://github.com/deis/postgres/issues)
-    - [helm classic](https://github.com/helm/helm-classic/issues)
-    - [monitor](https://github.com/deis/monitor/issues)
-    - [registry](https://github.com/deis/registry/issues)
-    - [router](https://github.com/deis/router/issues)
-    - [workflow](https://github.com/deis/workflow/issues)
-    - [workflow-cli](https://github.com/deis/workflow-cli/issues)
-- [Deis #community Slack channel][slack]
+Our official support channel is the [Deis #community Slack channel][slack].
 
 Don't know what the applicable repository for an issue is? Open up in issue in [workflow][] or chat with a maintainer in the [Deis #community Slack channel][slack] and we'll make sure it gets to the right place.
 
--- a/src/installing-workflow/configuring-object-storage.md
+++ b/src/installing-workflow/configuring-object-storage.md
@@ -11,7 +11,7 @@ Every component that relies on object storage uses two inputs for configuration:
 1. Component-specific environment variables (e.g. `BUILDER_STORAGE` and `REGISTRY_STORAGE`)
 2. Access credentials stored as a Kubernetes secret named `objectstorage-keyfile`
 
-The helm classic chart for Deis Workflow can be easily configured to connect Workflow components to off-cluster object storage. Deis Workflow currently supports Google Compute Storage, Amazon S3, Azure Blob Storage and OpenStack Swift Storage.
+The helm chart for Deis Workflow can be easily configured to connect Workflow components to off-cluster object storage. Deis Workflow currently supports Google Compute Storage, Amazon S3, Azure Blob Storage and OpenStack Swift Storage.
 
 ### Step 1: Create storage buckets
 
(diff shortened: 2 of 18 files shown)


src/contributing/overview.md
```diff
@@ -6,19 +6,7 @@ Interested in contributing to a Deis project?  There are lots of ways to help.
...
-    - [chart](https://github.com/deis/charts/issues)
-    - [database](https://github.com/deis/postgres/issues)
-    - [helm classic](https://github.com/helm/helm-classic/issues)
-    - [monitor](https://github.com/deis/monitor/issues)
-    - [registry](https://github.com/deis/registry/issues)
-    - [router](https://github.com/deis/router/issues)
-    - [workflow](https://github.com/deis/workflow/issues)
-    - [workflow-cli](https://github.com/deis/workflow-cli/issues)
-- [Deis #community Slack channel][slack]
+Our official support channels is the [Deis #community Slack channel][slack].
```
vdice: s/channels/channel

src/installing-workflow/configuring-registry.md
```diff
@@ -11,57 +11,21 @@ Every component that relies on registry uses two inputs for configuration:
...
-The helm classic chart for Deis Workflow can be easily configured to connect Workflow components to off-cluster registry. Deis Workflow supports external registries which provide either short lived tokens which are valid only for a specified amount of time or long lived tokens(basic username/password) which are valid forever for authenticating to them. For those registries which provide short lived tokens for authentication Deis Workflow will generate and refresh them such that the deployed apps will only have access to the short lived tokens and not to the actual credentials for the registries.
+The helm chart for Deis Workflow can be easily configured to connect Workflow components to off-cluster registry. Deis Workflow supports external registries which provide either short lived tokens which are valid only for a specified amount of time or long lived tokens(basic username/password) which are valid forever for authenticating to them. For those registries which provide short lived tokens for authentication Deis Workflow will generate and refresh them such that the deployed apps will only have access to the short lived tokens and not to the actual credentials for the registries.
```
vdice: (realizing none of this was actually modified in this PR, but...) 
1. could use a space between `lived tokens` and `(basic username/password)`
2. could use a comma between `...which provide short lived tokens for authentication` and `Deis Workflow will generate...`
3. properly, `short lived` should be `short-lived`

src/quickstart/provider/aws/install-aws.md
```diff
@@ -1,76 +1,45 @@
...
-    	$ helm repo add deis https://charts.deis.com/workflow  # add the workflow charts repo
-
-    	$ helm install deis/workflow --version=v2.8.0 --namespace=deis -f <optional values file>  # injects resources into your cluster
-
-  See also our section on [Workflow chart provenance](../../../installing-workflow/workflow-helm-charts.md#chart-provenance)
-
-Helm Classic will install a variety of Kubernetes resources in the `deis` namespace.
-You'll need to wait for the pods that it launched to be ready. Monitor their status
-by running:
+Helm will install a variety of Kubernetes resources in the `deis` namespace.
```
vdice: A note here that applies to all mentions of `helm install deis/workflow` -- I believe it is mandatory to add `--namespace deis` to said command for the chart to be installed into the `deis` namespace (or else, `default` will be used)

vdice:
Also wondering what we should do with https://github.com/deis/workflow/blob/master/src/installing-workflow/workflow-helm-charts.md ?  Should we transition it to be used for describing provenance (chart verification) specifically now that `helm` is the default elsewhere?

src/quickstart/provider/aws/install-aws.md
bacongobbler: Thanks for pointing that out! I'll update any references for that.

bacongobbler:
Hmm yeah, changing the title to how we sign charts and removing the Installation section sounds like a good idea.

src/installing-workflow/configuring-object-storage.md
```diff
@@ -25,172 +25,26 @@ If you provide credentials with sufficient access to the underlying storage, Wor
...
-| swift        | `SWIFT_USERNAME`, `SWIFT_PASSWORD`, `SWIFT_AUTHURL`, `SWIFT_AUTHVERSION`, `SWIFT_REGISTRY_CONTAINER`, `SWIFT_DATABASE_CONTAINER`, `SWIFT_BUILDER_CONTAINER` | To specify tenant set `SWIFT_TENANT` if the auth version is 2 or later.                             |
-
-!!! note
-	These environment variables should be set **before** running `helmc generate` in Step 5.
-
-**Option 2:** Using template file `tpl/generate_params.toml` available at `$(helmc home)/workspace/charts/workflow-v2.8.0`
-
-* Edit Helm Classic chart by running `helmc edit workflow-v2.8.0` and look for the template file `tpl/generate_params.toml` (make sure you have the `$EDITOR` environment variable set with your favorite text editor)
-* Update the `storage` parameter to reference the platform you are using, e.g. `s3`, `azure`, `gcs`, or `swift`
+* Fetch the Helm values by running `helm inspect values deis/workflow > values.yaml`
```
vdice: @bacongobbler I'm having trouble with this approach:
```
$ helm inspect values workflow/workflow > values.yaml

$ helm install workflow/workflow -f values.yaml --namespace=deis
Fetched workflow/workflow to workflow-v2.8.0.tgz
Error: error converting YAML to JSON: yaml: line 10: did not find expected <document start>

$ helm version
Client: &version.Version{SemVer:"v2.0.0", GitCommit:"e77d564b880c4436049f0723083d67ef810729b2", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.0.0", GitCommit:"51bdad42756dfaf3234f53ef3d3cb6bcd94144c2", GitTreeState:"clean"}
```
The reason seems to be that the `helm inspect` outputs the following line first: `Fetched workflow/workflow to workflow-v2.8.0.tgz` which is then the first line in `values.yaml`.  If this approach is pursued, that first line needs to be removed...

bacongobbler: Yep, I filed an issue for that in https://github.com/kubernetes/helm/issues/1574. Until that is addressed this workflow is bugged. Shall I add a note about that?

vdice: oh nice.  hmm, tough call on adding a note... it would be something we'd need to track and remove later...

mboersma: Since there probably won't be a new Helm release for a bit, we should add a note about removing the `Fetched workflow...` line.

bacongobbler: sure! the second line is a comment anyways so when this bug is gone we won't be affected. Sounds good to me!

bacongobbler:
done

PR #625, opened 2016-12-01
docs(installing-workflow): deprecate support for v1.2 clusters

bacongobbler:
We have not tested nor seen anyone use k8s v1.2 so I think it's fair to start dropping "official"
support for that minor release and just test against v1.3.4+ clusters.


bacongobbler:
[I stand corrected](https://github.com/deis/workflow/pull/625).

bacongobbler:
re-opening due to other issues like https://github.com/deis/builder/issues/455. Looks like Workflow silently made itself incompatible with v1.2 clusters. Since we actively test with v1.3.10 and v1.4.6 (and soon v1.5.0) let's deprecate v1.2 support.

PR #628, opened 2016-12-01
Update command used to auth to kubernetes host

chexxor:
Note that I'm a total newcomer to Deis. I also haven't verified this problem/fix from a technical point of view. It's worth sharing the fix here, though, in case it **is** verified and we want to fix it.

When I ran `$ kubectl cluster-info`, I got this error:
`error: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.`

Looks like the auth mechanism was updated a few months ago: https://github.com/kubernetes/kubernetes/issues/30617

> Previously, gcloud would have configured kubectl to use the cluster's static client certificate to authenticate. Now, gcloud is configuring kubectl to use the service account's credentials.
> 
> Kubectl is just using the Application Default Credentials library, and it looks like this is part of the ADC flow for using a JSON-key service account.

At the bottom of that thread, this command was suggested, which worked for me:
`$ gcloud auth application-default login`.

### DIFF
--- a/src/quickstart/provider/gke/boot.md
+++ b/src/quickstart/provider/gke/boot.md
@@ -72,11 +72,16 @@ If you haven't configured your default zone, make sure it matches the ZONE for y
 $ gcloud config set compute/zone us-central1-b
 ```
 
-Now you may fetch Kubernetes credentials:
+Now you may fetch credentials to connect to Kubernetes:
 ```
-$ gcloud container clusters get-credentials cluster-1
-Fetching cluster endpoint and auth data.
-kubeconfig entry generated for cluster-1.
+$ gcloud auth application-default login
+Your browser has been opened to visit:
+https://accounts.google.com/o/oauth2/auth?redirect_uri=....
+
+Credentials saved to file: [~/.config/gcloud/application_default_credentials.json]
+
+These credentials will be used by any library that requests
+Application Default Credentials.
 ```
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
Interesting. We're still using `gcloud container clusters get-credentials` as per https://cloud.google.com/sdk/gcloud/reference/container/clusters/get-credentials to fetch our account credentials. I'll see what I can find on the difference between the two.

bacongobbler:
Hmm, the thread you linked seems to summarize it quite well. LGTM!

bacongobbler:

chexxor:
Yeah, I'll try to do that today or tomorrow. Thanks for reviewing!

chexxor:
@bacongobbler, update. LMK if more changes needed.

bacongobbler:
Almost there! It's just missing a colon:

```
docs(gke/boot): update command to auth to kubernetes host
```

deisrel is a little finnicky :)

Thanks!

chexxor:
Would be nice to have a git hook for this. Perhaps "prepare-commit-msg" to put a default template in there, or "commit-msg" or "pre-push" to do a regex on the commit's first line. I haven't done much with hooks, but looks like you could commit them to the project in the "$GIT_DIR/hooks/*" directory.

bacongobbler:
running [`make commit-hook`](https://github.com/deis/controller/blob/master/Makefile#L39-L40) will give you what you desire. :)

PR #630, opened 2016-12-02
fix(managing-workflow): update logging and monitoring docs to reflect helm v2

### LINKED ISSUE #629: there is no helm generate in 2.0
There is no `helm generate` in 2.0. [Documentation](https://github.com/deis/workflow/search?utf8=%E2%9C%93&q=%22helm+generate%22) for Platform Logging and Monitoring doesn't point to the right directions anymore.

The best option to change something after the platform is installed (at least for my point) is:
- Run `helm install deis/workflow --namespace deis -f values.yaml --dry-run --debug | sed -e '1,/logger-fluentd-daemon.yaml/d;/---/,$d' > logger-fluentd-daemon.yaml`. 
- Add external log destination.
- Drop daemon set.
- Create new from the file. 

Other option might be using `kubectl get -o yaml`, but it includes headers which have to be removed anyway.

But, is it 'the best' way?

bacongobbler:
closes #629

### DIFF
--- a/src/managing-workflow/platform-logging.md
+++ b/src/managing-workflow/platform-logging.md
@@ -11,38 +11,20 @@ If the message is from the [Workflow Router](https://github.com/deis/router) we 
 Logger then acts as a consumer reading messages off of the NSQ logs topic storing those messages in a local Redis instance. When a user wants to retrieve log entries using the `deis logs` command we make an HTTP request from Controller to Logger which then fetches the appropriate data from Redis.
 
 ## Configuring Off Cluster Redis
-Even though we provide a redis instance with the default Workflow install. It is recommended that operators use a 3rd party source like Elasticache or similar offering. This way your data is durable across upgrades or outages. If you have a 3rd party Redis installation you would like to use all you need to do is set the following values in `generate_params.toml` within your chart's tpl directory.
+
+Even though we provide a redis instance with the default Workflow install, it is recommended that operators use a third-party source like Elasticache or similar offering. This way your data is durable across upgrades or outages. If you have a third-party Redis installation you would like to use all you need to do is set the following values in your helm chart:
 
 * db = "0"
 * host = "my.host.redis"
 * port = "6379"
 * password = ""
 
-You can also provide this environment variables when you run your `helm generate` command instead of editing `generate_params.toml`.
-
-* LOGGER_REDIS_LOCATION="off-cluster"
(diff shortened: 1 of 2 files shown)


szymonpk:
In [platform logging](https://github.com/deis/workflow/blob/master/src/managing-workflow/platform-logging.md#default-configuration), it's mentioned user should modify daemon's set configuration/manifest for external logging. However, there is no explanation how to do this, do we want to keep this in such way? Leave advanced users to discover this by themselves?

bacongobbler:
thanks @szymonpk. I've updated the documentation on how this would work with helm. Does that work for you?

szymonpk:
Now it's much closer to original documentation ;-) Thanks. I didn't knew about fetch with untar.

src/managing-workflow/platform-logging.md
```diff
@@ -11,20 +11,14 @@ If the message is from the [Workflow Router](https://github.com/deis/router) we
 Logger then acts as a consumer reading messages off of the NSQ logs topic storing those messages in a local Redis instance. When a user wants to retrieve log entries using the `deis logs` command we make an HTTP request from Controller to Logger which then fetches the appropriate data from Redis.
 
 ## Configuring Off Cluster Redis
-Even though we provide a redis instance with the default Workflow install. It is recommended that operators use a 3rd party source like Elasticache or similar offering. This way your data is durable across upgrades or outages. If you have a 3rd party Redis installation you would like to use all you need to do is set the following values in `generate_params.toml` within your chart's tpl directory.
+
+Even though we provide a redis instance with the default Workflow install, it is recommended that operators use a third-party source like Elasticache or similar offering. This way your data is durable across upgrades or outages. If you have a third-party Redis installation you would like to use all you need to do is set the following values in your chart:
```
vdice: Should we mention how the `logger redis` values can be set via `helm`?  (Something similar to [configuring object storage](https://github.com/deis/workflow/blob/687e916d0b955e467e6cb42263fe230f45560a39/src/installing-workflow/configuring-object-storage.md#step-4-configure-workflow-chart) perhaps?)

bacongobbler: Yeah, the values are listed below but it's quite vague. I'll see if I can clarify this better :)

bacongobbler:
I've added some more clarification on the process involved in a couple places in the platform logging docs by mentioning `helm fetch` and `helm inspect`.

PR #631, opened 2016-12-05
feat(charts): Add optional syslog endpoint setting to fluentd

rimusz:
PR https://github.com/deis/fluentd/pull/72

### DIFF
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -16,6 +16,7 @@ global:
   # - gcs: Store persistent data in Google Cloud Storage
   # - minio: Store persistent data on in-cluster Minio server
   storage: minio
+
   # Set the location of Workflow's PostgreSQL database
   #
   # Valid values are:
@@ -24,6 +25,7 @@ global:
   #   configured above)
   # - off-cluster: Run PostgreSQL outside the Kubernetes cluster (configure in database section)
   database_location: "on-cluster"
+
   # Set the location of Workflow's logger-specific Redis instance
   #
   # Valid values are:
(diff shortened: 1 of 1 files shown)


bacongobbler:
related: #630 

Depending on what's merged first, we'll have to update the documentation as well.

charts/workflow/values.yaml
```diff
@@ -130,6 +132,13 @@ logger:
     port: "redis port"
     password: "redis password" # "" == no password
 
+fluentd:
+  syslog:
+    # Configure the following ONLY if using Fluentd to send log messages to both
+    # the Logger component and external syslog endpoint
+    syslog_host: "external syslog endpoint url"
```
kmala: can you keep the values as empty like `syslog_host: ""`  and add the help test as comments above so nothing gets applied by default as https://github.com/deis/fluentd/pull/72/files?diff=unified#diff-65e48c77b3de178d48580127e51490c5R36 becomes true if the string is not empty.

charts/workflow/values.yaml
```diff
@@ -130,6 +132,15 @@ logger:
     port: "redis port"
     password: "redis password" # "" == no password
 
+fluentd:
+  syslog:
+    # Configure the following ONLY if using Fluentd to send log messages to both
+    # the Logger component and external syslog endpoint
+    # external syslog endpoint url
+    syslog_host: ""
```
bacongobbler: I think it should be `host` and `port` instead. `fluentd/syslog/syslog_host` is a little redundant since we're operating under the `syslog` namespace. I'm just bikeshedding at this point though, so LGTM.

bacongobbler:
I'll let someone else merge it if someone else feels my comment is invalid/bikeshedding at this point. :)

PR #641, opened 2016-12-08
fix(charts): fix workflow chart values

bacongobbler:
see deis/postgres#177
see deis/redis#10

### DIFF
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -113,22 +113,20 @@ controller:
   # admin_only - allows for registration by an admin only.
   registration_mode: "enabled"
 
-database:
-  # Configure the following ONLY if using an off-cluster PostgreSQL database
-  postgres:
-    name: "database name"
-    username: "database username"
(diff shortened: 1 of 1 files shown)


charts/workflow/values.yaml
```diff
@@ -113,22 +113,20 @@ controller:
...
-logger:
-  redis:
-    # Configure the following ONLY if using an off-cluster Redis instance for logger
-    db: "0"
-    host: "redis host"
-    port: "redis port"
-    password: "redis password" # "" == no password
+# Configure the following ONLY if using an off-cluster PostgreSQL database
+postgres:
+  name: "database name"
```
kmala: this doesn't work because the chart(secret) is under the database folder and this way the database chart can't access these values.

bacongobbler: Oh, so it's already namespaced by the chart name? That makes some sense... But why wouldn't off-cluster redis parameters get populated before deis/redis#10 then?

PR #643, opened 2016-12-08
fix(values): redis params are in the redis chart

bacongobbler:
THIS is the real fix.

closes deis/logger#127

### DIFF
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -122,13 +122,12 @@ database:
     host: "database host"
     port: "database port"
 
-logger:
-  redis:
-    # Configure the following ONLY if using an off-cluster Redis instance for logger
-    db: "0"
-    host: "redis host"
-    port: "redis port"
-    password: "redis password" # "" == no password
+redis:
+  # Configure the following ONLY if using an off-cluster Redis instance for logger
+  db: "0"
+  host: "redis host"
+  port: "redis port"
+  password: "redis password" # "" == no password
 
 monitor:
   grafana:


charts/workflow/values.yaml
```diff
@@ -122,13 +122,12 @@ database:
...
-    db: "0"
-    host: "redis host"
-    port: "redis port"
-    password: "redis password" # "" == no password
+redis:
+  # Configure the following ONLY if using an off-cluster Redis instance for logger
+  db: "0"
+  host: "redis host"
+  port: "redis port"
+  password: "redis password" # "" == no password
```
ineu: Looks like this is taken into account when logger's redis is set to 'on-cluster'. I get the following error:
```
*** FATAL CONFIG FILE ERROR ***
Reading the configuration file, at line 396
>>> 'requirepass redis password'
Bad directive or wrong number of arguments
```

bacongobbler: what version are you running? We haven't released a chart with this fix yet so it shouldn't occur with v2.9.0 or earlier. :)

ineu: I'm using 2.9.1:
```
> $ helm repo update                                                                                                                                                                          
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "stable" chart repository
...Successfully got an update from the "deis" chart repository
Update Complete. ⎈ Happy Helming!⎈ 
> $ helm search deis/workflow                                                                                                                                                                 
NAME         	VERSION	DESCRIPTION  
deis/workflow	v2.9.1 	Deis Workflow
> $ helm inspect deis/workflow | grep redis                                                                                                                                                   
  logger_redis_location: "on-cluster"
redis:
  host: "redis host"
  port: "redis port"
  password: "redis password" # "" == no password

```

ineu: Sure, will try to reproduce on a new cluster and open an issue if it succeeds

ineu: Cannot reproduce it at the moment. I think the reason of my issue was me changing redis from on-cluster to off-cluster and vice versa, probably some artifacts. Just set up 2.10 cluster, it runs with correct one-work `requirepass`.

PR #645, opened 2016-12-08
chore(chart): update values file description

vdice:
Description here is obviously minimal -- definitely open to further elaboration.  Main purpose was to get rid of existing description (copied over/mis-leading)


bacongobbler:
Works for me! As long as we keep one comment to work around https://github.com/kubernetes/helm/issues/1574 then we're good.

PR #662, opened 2016-12-15
docs(releases.md): add patch release notes

vdice:
Any other notes we should add inspired by the recent v2.9.1 release?

### DIFF
--- a/src/roadmap/releases.md
+++ b/src/roadmap/releases.md
@@ -61,6 +61,14 @@ deliverable. This section leads a maintainer through creating a component releas
 Major or minor releases should happen on the master branch. Patch releases
 should check out the previous release tag and cherry-pick specific commits from master.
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
One other note: you'll also need to manually kick off a `deis-com-deploy` job after pushing the branch with the cherry-picked commits. To clarify:

 - I created a `release-v2.9.1` branch in deis/workflow, based from the v2.9.0 tag
 - I cherry-picked the v2.9.1 release notes commit into the branch
 - I pushed the branch to deis/workflow
 - after approval, I tagged and pushed the v2.9.1 tag from the release-v2.9.1 branch
 - I manually ran `deis-com-deploy` job to ensure https://deis.com/docs/workflow/en/v2.9.1/ works (before it'd 404)

bacongobbler:
Either that or I could fix deis-com-deploy to be triggered on a pushed tag (probably better that way)

PR #668, opened 2016-12-20
docs(README.md): add temporary holiday notice

mboersma:
Let users know that the Workflow team may be temporarily short-handed over the holidays. ☃️


bacongobbler:
❄️

PR #673, opened 2016-12-28
chore(chart): remove empty default comment

vdice:
(Chart no longer inspects environment for IAM creds)

### DIFF
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -54,9 +54,9 @@ global:
 
 
 s3:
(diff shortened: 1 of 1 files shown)


vdice:
Note: I _may_ be mis-interpreting this comment... perhaps it is not alluding to reading env vars from host os but to the case where the k8s nodes are already set up with IAM perms (https://github.com/deis/workflow/blob/master/src/installing-workflow/configuring-object-storage.md#step-2-generate-storage-credentials) -- if so, I'll close.

PR #675, opened 2017-01-01
docs(applications) fix broken links for SSH keys

lenn4rd:
Links included the Markdown file extension.


bacongobbler:
Hmm, this seems like a mkdocs issue. The links are intended to have the .md extension as mkdocs will fix that when generated, but it looks like maybe the backslash is messing things up.

http://www.mkdocs.org/user-guide/writing-your-docs/#internal-hyperlinks

bacongobbler:
Try removing the backslashes instead and re-generate the docs locally; it should properly create the links as intended.

lenn4rd:
You were right, the slash breaks generating the link when an anchor is present. I removed it and the generated docs look fine now. These two files were the only ones affected.

PR #681, opened 2017-01-05
docs(*): add upgrade note for gcs/gcr

vdice:
Ref https://github.com/deis/registry-token-refresher/pull/7
Ref https://github.com/deis/workflow/pull/672

### DIFF
--- a/src/managing-workflow/upgrading-workflow.md
+++ b/src/managing-workflow/upgrading-workflow.md
@@ -31,6 +31,15 @@ period of downtime while the database recovers from backup.
 $ helm upgrade deis/workflow
 ```
 
+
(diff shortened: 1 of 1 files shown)


bacongobbler:
The code block doesn't render inside the admonition so it looks a bit awkward when reading it

2017-01-05-120802_3840x1080_scrot

Perhaps we should do away with the admonition here?

vdice:
@bacongobbler good point -- changed; let me know if that's what you had in mind.

mboersma:
I'm ok with it either way, honestly. Here is what it looks like after the latest update:
screen shot 2017-01-05 at 2 14 08 pm

bacongobbler:
Much better :)

PR #686, opened 2017-01-06
ref(upgrading-workflow): use -w 0 flag when base64 encoding

bacongobbler:
there's actually a flag for this :)


vdice:
@bacongobbler good call!  One more spot could use the simplification in https://github.com/deis/workflow/blob/master/src/installing-workflow/configuring-object-storage.md

bacongobbler:
done

PR #692, opened 2017-01-10
chore(roadmap): update product roadmap

bacongobbler:
Some things have been completed (such as deploy hooks) since the last roadmap update.

This commit also removes a few bullet points from the roadmap that have been dangling since v2.0
or have not been followed up with tickets/design docs (such as the controller emitting metrics).

cc'ing @sgoings @slack on whether we still want to create tickets for these enhancements.

### DIFF
--- a/src/roadmap/roadmap.md
+++ b/src/roadmap/roadmap.md
@@ -7,56 +7,6 @@ grouping of features that are deemed important to the future of Deis.
 Given the project's rapid [Release Schedule](releases.md), roadmap
 items are designed to provide a sense of direction over many releases.
 
-## Deployments
-
-Deis Workflow should use Kubernetes-native constructs wherever possible. By
-switching to `Deployments`, the controller component will no longer need to
-orchestrate rolling deploys. Instead, controller can delegate that work to
-Kubernetes control loops.
-
-* [X] Support Kubernetes Deployments <https://github.com/deis/controller/pull/854> (Workflow 2.2)
-* [X] Kubernetes Deployments by default (Workflow 2.4)
-
-## Private Registry and Native IaaS Registry Support
-
-Many users of Deis v1 requested the ability to push and pull application
-artifacts from private Docker registries. While v1 supported a
-[workaround](https://github.com/deis/deis/issues/2232) or two, operators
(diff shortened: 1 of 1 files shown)


bacongobbler:
Arguably "Trigger automatic deploy from GitHub Deployments API" is already done through the deis community-supported [dpl plugin](https://github.com/travis-ci/dpl) for deis.

mboersma:
> "Trigger automatic deploy from GitHub Deployments API" is already done

Is that dpl plugin specific to Travis CI in any way? If not, I agree we can cross that one off the list.

bacongobbler:
It is primarily used for TravisCI but there's nothing blocking someone from using it elsewhere.

Joshua-Anderson:
Triggering automatic deployments from the github api is very different from dpl. dpl is a command line tool that given some arguments can install dependencies and deploy a local git repo... `dpl --provider=deis --username=test --password=test --version=2.10.0 --app=test` ... installs the 2.10.0 deis client, auths the users, and runs git push. Travis CI uses this tool to run deployments after their CI finishes running.

Github deployments are an github api. A github user can trigger a deployment event on a github repository. Deployment tools can then listen to these events to trigger a deployment. In this case the deis controller would listen for github deployment events on a repo, and then will tell the builder to pull the specified version of the repo and deploy it.

bacongobbler:
Haha, thank you for schooling me, @Joshua-Anderson. You are indeed correct, as always. :)

What I meant to say is that the use case (deployments through git pushes to github) can be handled through travisCI via dpl. I believe @slack didn't intentionally mean that we should explicitly create a feature to write a feature around the Github Deployments API, just that users can deploy to Workflow when they push their code to Github.

As for

> Is that dpl plugin specific to Travis CI in any way? If not, I agree we can cross that one off the list.

We were just saying that someone could technically write a server that consumes the Github Deployments API and utilizes dpl to deploy to Workflow.

Joshua-Anderson:
Ah, I misinterpreted the scope of the feature. I'd be a huge fan of this feature, so I wanted to make sure it wasn't accidentally removed! But I see the point in it being some third-party community component rather than part of the core. Maybe if I have some (**gasp**) free time :)

PR #699, opened 2017-01-13
feat(charts): add optional persistence storage to influxdb and grafana

rimusz:


vdice:
@rimusz can you expand on the setup needed to run this?  I attempted to test locally (in combination with the monitor chart changes in https://github.com/deis/monitor/pull/169):
```
$ helm upgrade smelly-badger workflow-pr/workflow --version v2.7.1-20170119174902-sha.287dd60 -f workflow/values.yaml
Error: UPGRADE FAILED: unable to decode "": [pos 181]: json: error decoding base64 binary '<base64-encoded JSON data>': illegal base64 data at input byte 0
```
(where I use the workflow chart artifact that uses the monitor PR chart as its monitor sub-chart and I reference the local `workflow/values.yaml` to get the change in this PR)

As it seems I need to supply the `key_json`, I do so:
```
$ B64_KEY_JSON="$(cat /Users/vaughndice/.secrets/deis-sandbox-creds.json | base64 -w 0)"

$ helm upgrade smelly-badger workflow-pr/workflow --version v2.7.1-20170119174902-sha.287dd60 -f workflow/values.yaml --set gcs.key_json="${B64_KEY_JSON}",registry-token-refresher.gcr.key_json="${B64_KEY_JSON}"
```

...but then pods are stuck in CrashLoopBackOff.  I suppose my missing ingredient is specifically enabling use of GCP PD for my cluster?

rimusz:
ok, been testing myself on GKE and AWS (coreos) on k8s v1.4.x and v1.5.x and it looks like the `StorageClass` needs to be created before hand.
Not sure how to deal with that yet, as `StorageClass` is different for GCE/AWS

vdice:
@rimusz yes, I think a docs addition to this PR would be suitable in this case -- then, I can use those to guide testing.

Issue #700, opened 2017-01-14
Workflow helm install hangs if S3 bucket name has dots (.) in it


greglearns:
TL;DR: Workflow's helm chart should validate that the S3 database_bucket and registry_bucket do not have periods (.) in them.

THIS BREAKS THINGS

Modifying Deis Workflows basic chart values to include these changes break the deployment, and Workflow will hang on the deploy:
```
global:
  storage: s3

s3:
  # Your buckets.
  registry_bucket: "deis.subdomain.domain.com-registry"
  database_bucket: "deis.subdomain.domain.com-database"
  builder_bucket:  "deis.subdomain.domain.com-builder"
```
The builder bucket gets created in S3, but not the registry or database buckets.
The database logs include this at the end: 
```
ssl.CertificateError: hostname 'deis.subdomain.domain.com-registry.s3-us-west-2.amazonaws.com' doesn't match either of 's3-us-west-2.amazonaws.com', '*.s3-us-west-2.amazonaws.com', 's3.us-west-2.amazonaws.com', '*.s3.us-west-2.amazonaws.com', 's3.dualstack.us-west-2.amazonaws.com', '*.s3.dualstack.us-west-2.amazonaws.com', '*.s3.amazonaws.com'
2017/01/14 07:20:04 Error creating the registry bucket: exit status 1
```

THIS FIXES THINGS

Modifying Deis Workflows basic chart values to include these changes works, and Workflow will deploy:
```
global:
  storage: s3

s3:
  # Your buckets.
  registry_bucket: "deis-subdomain-domain-com-registry"
  database_bucket: "deis-subdomain-domain-com-database"
  builder_bucket:  "deis-subdomain-domain-com-builder"
```

Versions:

```
~/c/deis ❯❯❯ kubectl version
Client Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1", GitCommit:"82450d03cb057bab0950214ef122b67c83fb11df", GitTreeState:"clean", BuildDate:"2016-12-14T00:57:05Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"4", GitVersion:"v1.4.7", GitCommit:"92b4f971662de9d8770f8dcd2ee01ec226a6f6c0", GitTreeState:"clean", BuildDate:"2016-12-10T04:43:42Z", GoVersion:"go1.6.3", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #701, opened 2017-01-14
doc(installing-workflow): no periods in S3 bucket names

greglearns:
Modify configuring-object-storage.md to warn users not to use periods in S3 bucket names.

Related to https://github.com/deis/workflow/issues/700

### DIFF
--- a/src/installing-workflow/configuring-object-storage.md
+++ b/src/installing-workflow/configuring-object-storage.md
@@ -17,7 +17,7 @@ The helm chart for Deis Workflow can be easily configured to connect Workflow co
 
 Create storage buckets for each of the Workflow subsystems: `builder`, `registry`, and `database`.
 
-Depending on your chosen object storage you may need to provide globally unique bucket names.
+Depending on your chosen object storage you may need to provide globally unique bucket names. If you are using S3, use hyphens instead of periods in the bucket names. Using periods in the bucket name will cause an [ssl certificate validation issue with S3](https://forums.aws.amazon.com/thread.jspa?threadID=105357).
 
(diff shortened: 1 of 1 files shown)


bacongobbler:

greglearns:
@bacongobbler does that look better?

bacongobbler:
Perhaps could you elaborate in the docs as to why one should not use periods? It would help explain the reasoning behind this. :)

greglearns:
I put this in the commit message: Related to issue deis#700 https://github.com/deis/workflow/issues/700 because the explanation is a alittle long. I could copy that issue into this PR's commit message... is that what you are suggesting?

bacongobbler:
Longer explanations are great. This doc gets put up at http://deis.com/docs/workflow/installing-workflow/configuring-object-storage so nobody will be able to read the commit message.

greglearns:
I'm not sure of what is best to say. What about this change?
begin new text:
Depending on your chosen object storage you may need to provide globally unique bucket names. If you are using S3, do not use periods in the `registry_bucket` or `database_bucket` names (instead, use hyphens, if you wish); for example, `registry_bucket: "your.registry.bucket.name"` will fail, causing the affected service to not become Available.

bacongobbler:
Perhaps something like

>  Depending on your chosen object storage you may need to provide globally unique bucket names. If you are using S3, use hyphens instead of periods in the bucket names. Using periods in the bucket name will cause an [ssl certificate validation issue with S3](https://forums.aws.amazon.com/thread.jspa?threadID=105357).

greglearns:
Changed.

PR #703, opened 2017-01-17
docs(managing-workflow): Add link to terraform-provider-deis

olalonde:


bacongobbler:
From what I understand in https://github.com/deis/deis/issues/4824 this is only compatible with Deis v1 and is incomplete. Is this untrue?

bacongobbler:
Yes, this project is indeed only compatible with v1. Once it supports v2 then we can link it here. Let's close this for now until v2 support has been added. Thanks!

olalonde:
Ah ok, didn't notice that :(

PR #717, opened 2017-01-27
docs(api/v2.3): add TLS API examples and index v2.3 API

### LINKED ISSUE #650: Workflow Reference Guide not complete
The TLS rest calls are missing from the Workflow Reference Documentation.

/v2/apps/APP_NAME/tls

mboersma:
Closes #650

### DIFF
--- a/src/reference-guide/controller-api/v2.3.md
+++ b/src/reference-guide/controller-api/v2.3.md
@@ -576,6 +576,67 @@ DEIS_API_VERSION: 2.3
 DEIS_PLATFORM_VERSION: 2.3.0
 ```
 
+### Enable or disable TLS
+
+Example Request:
+
+```
+POST /v2/apps/example-go/tls/ HTTP/1.1
+Host: deis.example.com
+Content-Type: application/json
+Authorization: token abc123
+
+{
+  "https_enforced": true
+}
+```
+
+Example Response:
+
+```
+HTTP/1.1 201 CREATED
+DEIS_API_VERSION: 2.3
+DEIS_PLATFORM_VERSION: 2.3.0
+Content-Type: application/json
+
+{
+    "created": "2014-01-01T00:00:00UTC",
+    "app": "example-go",
+    "owner": "test",
+    "https_enforced": true,
+    "updated": "2014-01-01T00:00:00UTC",
+    "uuid": "de1bf5b5-4a72-4f94-a10c-d2a3741cdf75"
+}
+```
+
+### Get TLS status
+
+Example Request:
+
+```
+GET /v2/apps/example-go/tls/ HTTP/1.1
+Host: deis.example.com
+Authorization: token abc123
+```
+
+Example Response:
+
+```
+HTTP/1.1 200 OK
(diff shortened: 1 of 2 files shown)


src/reference-guide/controller-api/v2.3.md
```diff
@@ -576,6 +576,67 @@ DEIS_API_VERSION: 2.3
...
+Example Request:
+
+```
+POST /v2/apps/example-go/tls/ HTTP/1.1
+Host: deis.example.com
+Content-Type: application/json
+Authorization: token abc123
+
+{
+  "https_enabled": true
```
vdice: should the key be `https_enforced`?

src/reference-guide/controller-api/v2.3.md
```diff
@@ -576,6 +576,67 @@ DEIS_API_VERSION: 2.3
...
+HTTP/1.1 200 OK
+DEIS_API_VERSION: 2.3
+DEIS_PLATFORM_VERSION: 2.3.0
+Content-Type: application/json
+
+{
+    "created": "2014-01-01T00:00:00UTC",
+    "app": "example-go",
+    "owner": "test",
+    "https_enforced": false,
```
vdice: When I create a `foo` app, I'm seeing `"https_enforced": null`:
```
$ deis create foo --no-remote
Creating Application... done, created foo
If you want to add a git remote for this app later, use `deis git:remote -a foo`

$ curl ${CONTROLLER_HOST}/v2/apps/foo/tls/ -H "Authorization: token ${AUTH_TOKEN}" -H "Content-Type: application/json"
{"uuid":"c3abc1da-2b22-4ac7-8132-27574ced2332","app":"foo","owner":"v","created":"2017-01-27T17:49:03Z","updated":"2017-01-27T17:49:03Z","https_enforced":null}
```

(Using deis cli v2.10.0 and latest pre-release v2.10.1 dev chart)

mboersma: The server code effectively coerces `null` to `false`. Once you've explicitly disabled `deis tls` I guess it will say `false` there. Should we change the docs to say `null`?

vdice: Indeed, if `https_enforced` is explicitly set to `false` it is set as such.  It appears on app instantiation with no other actions, `https_enforced` is `null`/`not set`... No strong opinion for docs, just thought I'd mention.

src/reference-guide/controller-api/v2.3.md
bacongobbler: Indeed. See https://github.com/deis/controller/blob/f5d14548472b074ee0641e550a3cf92a194f353a/rootfs/api/models/tls.py#L11

src/reference-guide/controller-api/v2.3.md
bacongobbler: Yeah it's due to weird server behaviour. If we want to leave it unchanged when pushing application information, we send `null`. If we want to enable/disable, we send `true/false`, respectively.

PR #719, opened 2017-01-31
docs(aws): update for latest k8s version available

mboersma:
Some things have changed about `kube-up.sh`; this keeps the quick start for AWS current.


vdice:
followed/tested these updated docs and looks great!

PR #723, opened 2017-02-04
docs(src/applications/inter-app-communication.md) Add first draft

chexxor:
Here's a first draft at documenting communication between Deis apps, which would fix https://github.com/deis/workflow/issues/

I spent a lot of time trying to come up with a simple example of an http-client app which resolves an app's URL to its internal IP address. I couldn't find an existing docker image which runs properly on Deis, as I don't have a great understanding of what kind of docker image a Deis app can simply run. I found a "ping" image which successfully demonstrates DNS resolution, but the pings fail when run in Deis.

LMK of any other info which should be contained here. I'll probably take a second review later.

### DIFF
--- a/src/applications/inter-app-communication.md
+++ b/src/applications/inter-app-communication.md
@@ -1,0 +1,7 @@
+## Inter-app Communication
+
+A common architecture pattern of multi-process applications is to have one process serve public requests while having multiple other processes supporting the public one to, for example, perform actions on a schedule or process work items from a queue. To implement this system of apps in Deis Workflow, set up the apps to communicate using DNS resolution, as shown above, and hide the supporting processes from public view by removing them from the Deis Workflow router. See [Deis Blog: Private Applications on Workflow](https://deis.com/blog/2016/private-applications-on-deis-workflow/) for more details, which walks through an example of removing an app from the router.
+
+### DNS Service Discovery
+
+Deis Workflow supports deploying a single app composed of a system of processes. Each Deis Workflow app communicates on a single port, so communicating with another Workflow app means finding that app's address and port. All Workflow apps are mapped to port 80 externally, so finding its IP address is the only challenge. Workflow creates a [Kubernetes Service](https://kubernetes.io/docs/user-guide/services/) for each app, which effectively assigns a name and one cluster-internal IP address to an app. The DNS service running in the cluster adds and removes DNS records which point from the app name to its IP address as services are added and removed. Deis Workflow apps, then, can simply send requests to the domain name given to the service, which is "app-name.app-namespace".
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -49,6 +49,7 @@ pages:
     - Managing App Processes: applications/managing-app-processes.md
     - Managing App Configuration: applications/managing-app-configuration.md
     - Managing App Lifecycle: applications/managing-app-lifecycle.md
+    - Inter-app Communication: applications/inter-app-communication.md
     - Resource Limits: applications/managing-resource-limits.md
     - Domains and Routing: applications/domains-and-routing.md
     - SSL Certificates: applications/ssl-certificates.md


mboersma:
This is a good start on something that would be very useful to users. We should refer to "Workflow" or "Deis Workflow" consistently, instead of the old shorthand "Deis."

chexxor:
I've got a lot of PRs in many repos, so this one slipped off my radar. Thanks for the review!

chexxor:
Do we need to do something in mkdocs.yml to surface this new page? I was browsing this repo and came across that file.

bacongobbler:
Yes indeed, a change to mkdocs.yml to add it to the TOC will be required in order for this page to be available.

chexxor:
I've added the page to the mkdocs.yml.

Also, I've added a section to explain a closely related Deis Workflow feature: `deis routing:disable`. I want to add this so that I can link to the Deis blog post which served as great reference while creating this doc page.

vdice:
LGTM still stands -- good catch on mkdocs addition!

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
+## Inter-app Communication
+
+### DNS Service Discovery
+
+Deis Workflow supports deploying a single app composed of a system of processes. Each Deis Workflow app communicates on a single port, so communicating with another Workflow app means finding that app's address and port. All Workflow apps are mapped to port 80 externally, so finding its IP address is the only challenge. Workflow creates a [Kubernetes Service](https://kubernetes.io/docs/user-guide/services/) for each app, which effectively assigns a name and one cluster-internal IP address to an app. The DNS service running in the cluster adds and removes DNS records which point from the app name to its IP address as services are added and removed. Deis Workflow apps, then, can simply send requests to the domain name given to the service, which is "app-name.app-namespace".
+
+For example, if you deployed a Deis Workflow app like this:
+
+    $ deis create http-server --no-remote
+    $ deis pull deis/example-dockerfile-python:latest -a http-server
```
bacongobbler: example-go is usually the app of choice in the documentation, as the image is 10x as small as example-dockerfile-python:

```
><> docker images | grep example
deis/example-dockerfile-python                   latest                        811c39971f48        12 weeks ago        43.5 MB
deis/example-go                                  latest                        6d128cdcba34        2 years ago         4.072 MB
```

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
...
+
+You can deploy a second app and communicate with the first one:
+
+    $ deis create http-client --no-remote
+    $ # Need to set a port because the docker image doesn't use one.
+    $ deis config:set PORT=8080 -a http-client
+    $ deis config:set HOSTNAME=http-server.http-server
```
bacongobbler: You can set both of these in one line:

```
deis config:set PORT=8080 HOSTNAME=http-server.http-server -a http-client
```

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
...
+    1 packets transmitted, 0 packets received, 100% packet loss
+    ping http-server.http-server every 300 sec
+    PING http-server.http-server (10.71.241.103): 56 data bytes
+    
+    $ # While the ping failed, it successfully resolved
+    $ #  "http-server.http-server" to "10.71.241.103"
+
+Clean up:
+
+    $ deis destroy -a http-server --confirm=http-server
```
bacongobbler: I don't believe cleanup should be documented. The user following the guide may want to play around with the example some more.

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
...
+    $ deis create http-server --no-remote
+    $ deis pull deis/example-dockerfile-python:latest -a http-server
+
+You can deploy a second app and communicate with the first one:
+
+    $ deis create http-client --no-remote
+    $ # Need to set a port because the docker image doesn't use one.
+    $ deis config:set PORT=8080 -a http-client
+    $ deis config:set HOSTNAME=http-server.http-server
+    $ deis pull -a http-client willfarrell/ping:latest
```
bacongobbler: Running `deis pull` on this image fails because it does not listen on port 8080 and fails the health check.

```
><> deis pull -a http-client willfarrell/ping:latest
Creating build... Error: Unknown Error (400): {"detail":"(app::deploy): Readiness probe failed: dial tcp 172.17.0.26:8080: getsockopt: connection refused"}
```

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
...
+
+Clean up:
+
+    $ deis destroy -a http-server --confirm=http-server
+    $ deis destroy -a http-client --confirm=http-client
+
+
+### Environment Variable Service Discovery
+
+It's useful to note that Kubernetes also supports [environment variable service discovery](https://kubernetes.io/docs/user-guide/services/#environment-variables), which works by adding environment variables to an app's environment to provide a reference to other services in the app's namespace. This method is currently not supported by Deis Workflow because Workflow creates a separate namespace for each app it deploys.
```
bacongobbler: We shouldn't document this if we don't plan on supporting it. Either this should reference a ticket like https://github.com/deis/controller/issues/710 to make an action item or should be removed completely.

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
...
+
+### Related Topics
+
+A common architecture pattern of multi-process applications is to have one process serve public requests while having multiple other processes supporting the public one to, for example, perform actions on a schedule or process work items from a queue. To implement this system of apps in Deis Workflow, set up the apps to communicate using DNS resolution, as shown above, and hide the supporting processes from public view by removing them from the Deis Workflow router.
```
bacongobbler: I think this should go up further as a supporting use case for inter-app communication rather than a separate heading.

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
...
+
+    deis routing:disable -a supporting-scheduled-app
```
bacongobbler: missing $ character to note this is a console command.

src/applications/inter-app-communication.md
```diff
@@ -0,0 +1,46 @@
+## Inter-app Communication
+
+### DNS Service Discovery
+
+Deis Workflow supports deploying a single app composed of a system of processes. Each Deis Workflow app communicates on a single port, so communicating with another Workflow app means finding that app's address and port. All Workflow apps are mapped to port 80 externally, so finding its IP address is the only challenge. Workflow creates a [Kubernetes Service](https://kubernetes.io/docs/user-guide/services/) for each app, which effectively assigns a name and one cluster-internal IP address to an app. The DNS service running in the cluster adds and removes DNS records which point from the app name to its IP address as services are added and removed. Deis Workflow apps, then, can simply send requests to the domain name given to the service, which is "app-name.app-namespace".
+
+For example, if you deployed a Deis Workflow app like this:
```
bacongobbler: Overall I think the documentation is going in the right direction, but I don't think the demonstration is useful. Perhaps just remove the demo and leave the last sentence?

> The [Deis Blog: Private Applications on Workflow](https://deis.com/blog/2016/private-applications-on-deis-workflow/) provides more details by walking through an example of internal cluster DNS communication.

bacongobbler:
ping @chexxor, will you be able to come back to this, or shall I remove this from the v2.13.0 milestone?

src/applications/inter-app-communication.md
chexxor: It's my understanding that Deis wants to follow Heroku's model, and Heroku doesn't support separate namespaces for each. That issue doesn't indicate any desire by core team to support defining a namespace. It's better to document this design decision than make people wonder.

src/applications/inter-app-communication.md
chexxor: I disagree. The kubernetes tutorial docs demonstrate cleanup [hello-minikube](https://kubernetes.io/docs/tutorials/stateless-application/hello-minikube/). Also, this command and its comment are pretty clear, so a user wanted to continue playing can simply wait to execute this until they are done playing. Or they can restart from the beginning to start playing again.

src/applications/inter-app-communication.md
chexxor: I need a better suggestion if I'm going to change it. I spent a lot of time thinking about how to demonstrate inter-app comms, and this is the best I could find.

src/applications/inter-app-communication.md
chexxor: What isn't useful about the demonstration? Can we find a better one?

bacongobbler: The blog post I linked has two working apps that communicate internally using DNS.

src/applications/inter-app-communication.md
bacongobbler: https://github.com/bacongobbler/nginx-upstream

src/applications/inter-app-communication.md
chexxor: The blog post is nice, but the http client is more complicated than I like. It requires knowledge of nginx, proxy, and authentication. This example only requires knowledge of unix ping, which is more simple, IMO.

Also, a blog post is different than a doc page. Linking to a blog post isn't an acceptable form of docs, IMO. For example, if CLI tool changes, then we can't grep the doc repo to know there's a doc page we need to update.

src/applications/inter-app-communication.md
chexxor: I get a readiness check failure when using deis/example-go.

```
$ deis pull deis/example-go -a http-server
Creating build... Error: Unknown Error (400): {"detail":"(app::deploy): Readiness probe failed: dial tcp 10.68.7.165:8080: getsockopt: connection refused"}
```

Same error when pulling "deis/example-dockerfile-python:latest", so I dunno how to fix.

chexxor:
Updated. Thanks for review!

bacongobbler:

chexxor:
Thanks for maintaining!

bacongobbler:
No, thank you.

PR #732, opened 2017-02-09
feat(ingress) Experimental Native Ingress

kris-nova:
Work for experimental native ingress. Run the documentation to see how to use the new feature.

Requires deis/controller#1243
Requires deis/builder#495
Requires https://github.com/deis/router/pull/316

### DIFF
--- a/charts/workflow/requirements.yaml
+++ b/charts/workflow/requirements.yaml
@@ -45,7 +45,7 @@ dependencies:
     version: <registry-token-refresher-tag>
     repository: https://charts.deis.com/registry-token-refresher
   - name: router
-    condition: !global.experimental_native_engress
+    condition: !global.experimental_native_ingress
     version: <router-tag>
     repository: https://charts.deis.com/router
   - name: workflow-manager
--- a/charts/workflow/requirements.yaml
+++ b/charts/workflow/requirements.yaml
@@ -45,7 +45,6 @@ dependencies:
     version: <registry-token-refresher-tag>
     repository: https://charts.deis.com/registry-token-refresher
   - name: router
-    condition: !global.experimental_native_ingress
     version: <router-tag>
     repository: https://charts.deis.com/router
   - name: workflow-manager
--- a/charts/workflow/requirements.yaml
+++ b/charts/workflow/requirements.yaml
@@ -45,6 +45,7 @@ dependencies:
     version: <registry-token-refresher-tag>
     repository: https://charts.deis.com/registry-token-refresher
   - name: router
+    condition: !global.experimental_native_engress
     version: <router-tag>
     repository: https://charts.deis.com/router
   - name: workflow-manager
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -51,11 +51,11 @@ global:
   host_port: 5555
   # Prefix for the imagepull secret created when using private registry
   secret_prefix: "private-registry"
-  # Experimental feature to toggle using kubernetes ingress instead of the Deis router.
+  # Experimental feature to use Kubernetes ingress instead of Workflow's deis-router.
   #
   # Valid values are:
-  # - true: The Deis router will NOT be deployed. Inherently workflow will not be usable until a Kubernetes ingress controller is installed.
-  # - false: The default mode, and the default behavior of Deis workflow.
+  # - true: deis-router will not be deployed. Workflow will not be usable until a Kubernetes ingress controller is installed.
+  # - false: deis-router will be deployed (default).
   experimental_native_ingress: false
 
 
@@ -113,7 +113,7 @@ controller:
   # disabled - turns off open registration
   # admin_only - allows for registration by an admin only.
   registration_mode: "admin_only"
-  # The public resolvable hostname to build your cluster with.
+  # The publicly resolvable hostname to build your cluster with.
   #
   # This will be the hostname that is used to build endpoints such as "deis.$HOSTNAME"
   platform_domain: ""
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -51,10 +51,6 @@ global:
   host_port: 5555
   # Prefix for the imagepull secret created when using private registry
   secret_prefix: "private-registry"
-  # The public resolvable hostname to build your cluster with.
-  #
-  # This will be the hostname that is used to build endpoints such as "deis.$HOSTNAME"
-  hostname: ""
   # Experimental feature to toggle using kubernetes ingress instead of the Deis router.
   #
   # Valid values are:
@@ -117,7 +113,11 @@ controller:
   # disabled - turns off open registration
   # admin_only - allows for registration by an admin only.
   registration_mode: "admin_only"
-
+  # The public resolvable hostname to build your cluster with.
+  #
+  # This will be the hostname that is used to build endpoints such as "deis.$HOSTNAME"
+  platform_domain: ""
+  
 database:
   # The username and password to be used by the on-cluster database.
   # If left empty they will be generated using randAlphaNum
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -51,6 +51,16 @@ global:
   host_port: 5555
   # Prefix for the imagepull secret created when using private registry
   secret_prefix: "private-registry"
+  # The public resolvable hostname to build your cluster with.
+  #
+  # This will be the hostname that is used to build endpoints such as "deis.$HOSTNAME"
+  hostname: ""
+  # Experimental feature to toggle using kubernetes ingress instead of the Deis router.
+  #
+  # Valid values are:
+  # - true: The Deis router will NOT be deployed. Inherently workflow will not be usable until a Kubernetes ingress controller is installed.
+  # - false: The default mode, and the default behavior of Deis workflow.
+  experimental_native_ingress: false
 
 
 s3:
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -117,7 +117,7 @@ controller:
   #
   # This will be the hostname that is used to build endpoints such as "deis.$HOSTNAME"
   platform_domain: ""
-  
+
 database:
   # The username and password to be used by the on-cluster database.
   # If left empty they will be generated using randAlphaNum
--- a/src/installing-workflow/experimental-native-ingress.md
+++ b/src/installing-workflow/experimental-native-ingress.md
@@ -8,9 +8,7 @@ Now that Helm is installed and the repository has been added, install Workflow w
 $ helm install deis/workflow --namespace deis --set global.experimental_native_ingress=true,controller.platform_domain=deis.com
 ```
 
-Where `controller.platform_domain` is a **required** parameter that is traditionally not required for Workflow. In this example we are using `deis.com` for `$hostname`.
-
-
+Where `controller.platform_domain` is a **required** parameter that is traditionally not required for Workflow that is explained in the next section. In this example we are using `deis.com` for `$hostname`.
  
 Helm will install a variety of Kubernetes resources in the `deis` namespace.
 Wait for the pods that Helm launched to be ready. Monitor their status by running:
@@ -25,12 +23,11 @@ You should also notice that several Kubernetes ingresses has been installed on y
 $ kubectl get ingress --namespace deis
 ```
 
-
 Depending on the order in which the Workflow components initialize, some pods may restart. This is common during the
 installation: if a component's dependencies are not yet available, that component will exit and Kubernetes will
 automatically restart it.
 
-Here, it can be seen that the controller, builder and registry all took a few loops before they were able to start:
+Here, it can be seen that the controller, builder and registry all took a few loops waiting for minio before they were able to start:
 
 ```
 $ kubectl --namespace=deis get pods
--- a/src/installing-workflow/experimental-native-ingress.md
+++ b/src/installing-workflow/experimental-native-ingress.md
@@ -48,7 +48,7 @@ deis-workflow-manager-68nu6   1/1       Running   0          5m
 
 ## Install a Kubernetes Ingress Controller
 
-Now that Workflow has been deployed with the `global.exerpimental_native_ingress` flag set to `true`, we will need a Kubernetes ingress controller in place to begin routing traffic.
+Now that Workflow has been deployed with the `global.experimental_native_ingress` flag set to `true`, we will need a Kubernetes ingress controller in place to begin routing traffic.
 
 Here is an example of how to use [traefik](https://traefik.io/) as an ingress controller for Workflow. Of course, you are welcome to use any controller you wish.
 
--- a/src/installing-workflow/experimental-native-ingress.md
+++ b/src/installing-workflow/experimental-native-ingress.md
@@ -8,7 +8,7 @@ Now that Helm is installed and the repository has been added, install Workflow w
 $ helm install deis/workflow --namespace deis --set global.experimental_native_ingress=true,controller.platform_domain=deis.com
 ```
 
-Where `global.hostname` is a **required** parameter that is traditionally not required for Workflow. In this example we are using `deis.com` for `$hostname`.
+Where `controller.platform_domain` is a **required** parameter that is traditionally not required for Workflow. In this example we are using `deis.com` for `$hostname`.
 
 
  
@@ -19,7 +19,7 @@ Wait for the pods that Helm launched to be ready. Monitor their status by runnin
 $ kubectl --namespace=deis get pods
 ```
 
-You should also notice that a Kubernetes ingress has been installed on your cluster. You can view it by running:
+You should also notice that several Kubernetes ingresses has been installed on your cluster. You can view it by running:
 
 ```
 $ kubectl get ingress --namespace deis
@@ -60,7 +60,7 @@ $ helm install stable/traefik --name deis-ingress-001 --namespace kube-system
 
 The experimental ingress feature requires a user to set up a hostname, and assumes the `deis.$host` convention.
 
-We need to point the `deis.$host` record to the public IP address of your ingress controller. You can get the public IP using the following command.
+We need to point the `*.$host` record to the public IP address of your ingress controller. You can get the public IP using the following command. A wildcard entry is necessary here as apps will use the same rule after they are deployed.
 
 ```
 $ kubectl get svc deis-ingress-001 --namespace kube-system
@@ -78,3 +78,7 @@ If we were using `deis.com` as a hostname we would need to create the following 
 Once all of the pods are in the `READY` state, and `deis.$host` resolves to the external IP found above Workflow is up an running!
 
 After installing Workflow, [register a user and deploy an application](../quickstart/deploy-an-app.md).
+
+##### Feedback
+
+While this feature is experimental we welcome feedback on the issue. We would like to learn more about use cases, and user experience. Please [open a new issue](https://github.com/deis/workflow/issues/new) for feedback.
--- a/src/installing-workflow/experimental-native-ingress.md
+++ b/src/installing-workflow/experimental-native-ingress.md
@@ -5,7 +5,7 @@
 Now that Helm is installed and the repository has been added, install Workflow with a native ingress by running:
 
 ```
-$ helm install deis/workflow --namespace deis --set experimental_native_ingress=true,global.hostname="deis.com"
+$ helm install deis/workflow --namespace deis --set global.experimental_native_ingress=true,controller.platform_domain=deis.com
 ```
 
 Where `global.hostname` is a **required** parameter that is traditionally not required for Workflow. In this example we are using `deis.com` for `$hostname`.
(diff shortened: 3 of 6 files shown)


.gitignore
```diff
@@ -3,6 +3,7 @@
 # Packages
 *.egg
 *.egg-info
+.idea*
```
mboersma: I'm a broken record, sorry, but ideally this would be in a separate PR.

charts/workflow/values.yaml
```diff
@@ -51,6 +51,16 @@ global:
   host_port: 5555
   # Prefix for the imagepull secret created when using private registry
   secret_prefix: "private-registry"
+  # The public resolvable hostname to build your cluster with.
+  #
+  # This will be the hostname that is used to build endpoints such as "deis.$HOSTNAME"
+  hostname: ""
```
mboersma: What happens if `experimental_native_ingress: true`, but the `hostname` field here wasn't updated? Ideally that would lead to a hard-stop error before `helm` actually installs the chart, but I don't know off-hand how to accomplish that.

krancour: This is analogous to the router chart's `platform_domain` setting, yes?

Can I propose that `platform_domain` is a better name than `hostname`? And to both 1. avoid confusion with the setting of the same name under `router` and 2. move this configuration closer to the point of use, this belongs under `controller` settings instead of `global` (because, afaik, the controller chart, which will have to honor this new setting won't be able to "look up" to the "parent" chart's global settings).

kris-nova: @mboersma As it stands we will get a regex error **before** we install the chart.

```
kris-nova:workflow kris$ helm install . --namespace deis --set global.experimental_native_ingress=true
Error: release elegant-cardinal failed: Ingress.extensions "controller-api-server-ingress" is invalid: spec.rules[0].host: Invalid value: "deis.": must match the regex [a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)* (e.g. 'example.com')
kris-nova:workflow kris$ 
```

kris-nova:
Example command for reference:

```
 helm install . --namespace deis --set global.experimental_native_ingress=true,controller.platform_domain=fabulous.af
```

charts/workflow/values.yaml
kris-nova: @krancour thanks for the suggestions, I changed `hostname` to `platform_domain` and pulled it into the controller and out of global.

src/quickstart/deploy-an-app.md
```diff
@@ -1,26 +1,44 @@
-## Register an Admin User
+## Determine your host and hostname values
 
-The first user to register against Deis Workflow will automatically be given administrative privileges.
+For the rest of this example we will refer to a special variables called `$hostname`. Please choose one of the two methods for building your `$hostname`.
 
-If you installed Deis on GKE or AWS, Deis automatically creates a load balancer for the cluster. To get the IP of this load balancer, run `kubectl --namespace=deis describe svc deis-router`.
+#### Option 1: Standard Installation 
+
+For a standard installation you can build the hostname using public IP address and a wildcard DNS solution. Instead of setting up DNS records, this example will use `nip.io`.
```
mboersma: Maybe:

"For a standard installation that includes deis-router, you can build the hostname using its public IP address and a wildcard DNS record."

src/quickstart/deploy-an-app.md
```diff
@@ -1,26 +1,44 @@
...
 If you do not have an load balancer IP, the router automatically forwards traffic from a kubernetes node to the router. In this case, use the IP of a kubernetes node and the node
 port that routes to port 80 on the controller.
 
-Deis requires a wildcard DNS record to dynamically map app names to the router. Instead of setting up DNS records, this example will use `nip.io`. If your router IP is `1.1.1.1`, its url will be `1.1.1.1.nip.io`. The URL of the controller component will be `deis.1.1.1.1.nip.io`.
+Deis requires a wildcard DNS record to dynamically map app names to the router.
+
+#### Option 2: Experimental Native Ingress Installation
 
-Use the controller url to register a user in the cluster.
+In this example, the user should already have DNS set up pointing to their known host. The `$hostname` value can be build by appending `deis.` to the value set in `global.exerpimental_native_ingress`.
```
mboersma: This should say "prepending" right?

src/quickstart/deploy-an-app.md
```diff
@@ -1,26 +1,44 @@
...
+ 
+```
+kubectl --namespace=deis describe svc deis-router
+```
 
 If you do not have an load balancer IP, the router automatically forwards traffic from a kubernetes node to the router. In this case, use the IP of a kubernetes node and the node
 port that routes to port 80 on the controller.
 
-Deis requires a wildcard DNS record to dynamically map app names to the router. Instead of setting up DNS records, this example will use `nip.io`. If your router IP is `1.1.1.1`, its url will be `1.1.1.1.nip.io`. The URL of the controller component will be `deis.1.1.1.1.nip.io`.
+Deis requires a wildcard DNS record to dynamically map app names to the router.
```
mboersma: "Deis" -> "Deis Workflow"

src/quickstart/deploy-an-app.md
mboersma: Typo: "exerpimental"

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Install Deis Workflow (With experimental native ingress support)
+
+Now that Helm is installed and the repository has been added, install Workflow with a native ingress by running:
+
+```
+$ helm install deis/workflow --namespace deis --set global.experimental_native_ingress=true,controller.platform_domain=deis.com
+```
+
+Where `global.hostname` is a **required** parameter that is traditionally not required for Workflow. In this example we are using `deis.com` for `$hostname`.
```
mboersma: Are we using `global.hostname` or `controller.platform_domain`? The example disagrees with this sentence.

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,80 @@
...
+
+
+ 
+Helm will install a variety of Kubernetes resources in the `deis` namespace.
+Wait for the pods that Helm launched to be ready. Monitor their status by running:
+
+```
+$ kubectl --namespace=deis get pods
```
mboersma: add `-w`?

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+You should also notice that a Kubernetes ingress has been installed on your cluster. You can view it by running:
```
mboersma: "a Kubernetes ingress *controller* has been installed"?

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,80 @@
...
+deis-logger-fluentd-esm60     1/1       Running   0          5m
+deis-logger-sm8b3             1/1       Running   0          5m
+deis-minio-4ww3t              1/1       Running   0          5m
+deis-registry-asozo           1/1       Running   1          5m
+deis-workflow-manager-68nu6   1/1       Running   0          5m
+```
+
+## Install a Kubernetes Ingress Controller
+
+Now that Workflow has been deployed with the `global.exerpimental_native_ingress` flag set to `true`, we will need a Kubernetes ingress controller in place to begin routing traffic.
```
mboersma: Typo: "exerpimental"

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,80 @@
...
+If we were using `deis.com` as a hostname we would need to create the following A DNS record.
+
+| Name              | Type          | Value           |
+| ----------------- |:-------------:| ---------------:|
+| deis.deis.com     | A             | 104.154.159.184 |
+
+
+Once all of the pods are in the `READY` state, and `deis.$host` resolves to the external IP found above Workflow is up an running!
+
+After installing Workflow, [register a user and deploy an application](../quickstart/deploy-an-app.md).
```
mboersma: Since it's experimental, should we provide a link in the docs to github issue reporting? Or encourage users to share their stories of success with different ingress controllers by submitting docs PRs as HOWTOs?

Not necessary, just thinking...

charts/workflow/values.yaml
```diff
@@ -51,6 +51,12 @@ global:
   host_port: 5555
   # Prefix for the imagepull secret created when using private registry
   secret_prefix: "private-registry"
+  # Experimental feature to toggle using kubernetes ingress instead of the Deis router.
```
rimusz: indeed, it confuses people

rimusz:
What is missing I see is the ingress rules annotation e.g. `kubernetes.io/ingress.class: "traefik"` setup option in controller/apps ingress rules. Without it the traffic will not pass ingress controller but each rule will get the loadbalancer created instead

krancour:
@rimusz @kris-nova 

> What is missing I see is the ingress rules annotation e.g. kubernetes.io/ingress.class: "traefik" setup option in controller/apps ingress rules.

Well... as much as I love it. Let's not lock ourselves into Traefik. More on that in a second.

> Without it the traffic will not pass ingress controller but each rule will get the loadbalancer created instead.

Yes... because that's how the L7 ingress controller (which is a default in GKE behaves). This isn't _wrong_. It just might not be what you want. As you've alluded to, various annotations are typically used as signals to different ingress controllers that they either _should_ or should _not_ act on any given ingress.

@kris-nova buried somewhere in the Google doc I'd written up about this, this detail was discussed. I probably should have called more attention to it. What we will need to do is allow, within this chart, for users to specify annotations as k/v pairs in their `values.yaml`. Those will have to be passed to the controller either through an env var or configmap and the controller will have to ensure the indicated annotations are added to each ingress it creates.

(If that's already present, I apologize. I am just now starting to review in earnest, and I am taking a guess that @rimusz wouldn't have overlooked this if it were already there.)

> The case above does not apply for AWS, but for GKE/GCP is needed, as otherwise the GKE/GCP controller will fight the custom controller for the Ingress.

Right. Because no solution (that I am familiar with) for provisioning k8s on AWS includes a default ingress controller of any kind. So in _practical_ terms, this only (currently) is a problem in GKE, but it can become a problem _anywhere_ if someone has (for whatever reasons) installed more than one ingress controller.

charts/workflow/requirements.yaml
```diff
@@ -45,6 +45,7 @@ dependencies:
     version: <registry-token-refresher-tag>
     repository: https://charts.deis.com/registry-token-refresher
   - name: router
+    condition: !global.experimental_native_ingress
```
krancour: Does the `!` operator actually work here? I was hoping it would, but I wasn't sure and hadn't had a chance to verify yet.

krancour: But if it works... this is _exactly_ what I had in ming for how the router chart, in its entirety, is conditionally included or excluded. I ❤️ this.

charts/workflow/values.yaml
```diff
@@ -107,6 +113,10 @@ controller:
   # disabled - turns off open registration
   # admin_only - allows for registration by an admin only.
   registration_mode: "admin_only"
+  # The public resolvable hostname to build your cluster with.
```
krancour: Even with the following line clarifying it, I somehow still feel thrown off by the phrase "to build your cluster with."

Can we maybe replace this and the following sentence with something like:

```
A resolvable domain or subdomain name. All applications created by the controller will, by default, be subdomains of this. When experimental native ingress is enabled, these subdomains are used in defining host-based routing rules. If, for instance, platform_domain were "example.com", for the application "foobar", the controller will create rules to route `foo.example.com` to the appropriate backend.
```

^ And that's totally imperfect. Feel free to iterate on that if you can improve it further.

src/installing-workflow/experimental-native-ingress.md
krancour: `controller.platform_domain`. It was changed elsewhere but not here.

src/installing-workflow/experimental-native-ingress.md
krancour: "ingress" is the proper terminology here, as @kris-nova is referring to ingress _resources_ that are part of the chart and define routing rules for the controller, grafana, etc.

That being said, "a Kubernetes ingress," might need to change because I do think there's more than one installed as part of the platform itself.

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,80 @@
+# Experimental Native Ingress
```
krancour: I'm wondering if this page repeats too much documentation from the "normal" doc for installing without this experimental feature. Do we want to consider a lesser page that just highlights what's different when you install with this feature flag set? I really don't know what's best here. I'm just asking the question.

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,80 @@
...
+
+```
+$ helm install stable/traefik --name deis-ingress-001 --namespace kube-system
+```
+
+## Configure DNS
+
+The experimental ingress feature requires a user to set up a hostname, and assumes the `deis.$host` convention.
+
+We need to point the `deis.$host` record to the public IP address of your ingress controller. You can get the public IP using the following command.
```
krancour: Shouldn't it actually be `*.$host` that resolves to their ingress controller?

src/quickstart/deploy-an-app.md
krancour: > ... to the value set in `global.exerpimental_native_ingress`.

Should that be `controller.platform_domain`?

charts/workflow/requirements.yaml
kris-nova: So I decided to take this functionality out for this PR.. seems out of scope, and honestly if it's critical that we move our abstraction out to another abstracted place of abstraction we can PR that later.

Also this is blocked on a `helm` issue: https://github.com/kubernetes/helm/issues/2111

charts/workflow/values.yaml
kris-nova: Changed to "_publicly resolvable_" to keep it simple

If we want to elaborate more on this later we can.

src/installing-workflow/experimental-native-ingress.md
kris-nova: So I already considered hooking into the existing page, I think it makes sense to have a clear change in flow for an entire section, than to try to add a bunch of conditional logic to an existing flow. I visualize the flows like this:

### Separate and clear
```
              ----
----- <           > ------
              ----
```

### Together and conditional

```
---- <>-<>-<>-<>
```

src/installing-workflow/experimental-native-ingress.md
kris-nova: Kubernetes will have the pod information stored **very** quickly, and I think adding a `-w` adds an implicit requirement on noting `^C` and that sounds like adding a lot for a small gain.

src/quickstart/deploy-an-app.md
kris-nova: Sorry to hear that @krancour, do you have a suggestion on how it could be better?

kris-nova: I went ahead and 

```
s/build the hostname/calculate the hostname value/g`
```

..if that helps?

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Install Deis Workflow (With experimental native ingress support)
+
+Now that Helm is installed and the repository has been added, install Workflow with a native ingress by running:
+
+```
+$ helm install deis/workflow --namespace deis --set global.experimental_native_ingress=true,controller.platform_domain=deis.com
+```
+
+Where `controller.platform_domain` is a **required** parameter that is traditionally not required for Workflow. In this example we are using `deis.com` for `$hostname`.
```
bacongobbler: Should we make a prelude here saying that this is documented further below in "Configuring DNS"?

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,84 @@
...
+
+
```
bacongobbler: there's a few extra unneeded newlines here inbetween paragraphs. Mind removing them?

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,84 @@
...
+```
+$ kubectl --namespace=deis get pods
+```
+
+You should also notice that several Kubernetes ingresses has been installed on your cluster. You can view it by running:
+
+```
+$ kubectl get ingress --namespace deis
+```
+
```
bacongobbler: extra newline here

src/installing-workflow/experimental-native-ingress.md
```diff
@@ -0,0 +1,84 @@
...
+
+Depending on the order in which the Workflow components initialize, some pods may restart. This is common during the
+installation: if a component's dependencies are not yet available, that component will exit and Kubernetes will
+automatically restart it.
+
+Here, it can be seen that the controller, builder and registry all took a few loops before they were able to start:
```
bacongobbler: took a few loops waiting for minio*

src/installing-workflow/index.md
```diff
@@ -33,6 +33,9 @@ More rigorous installations would benefit from using outside sources for the fol
 * [Redis](../managing-workflow/platform-logging.md#configuring-off-cluster-redis) - Such as AWS Elasticache
 * [InfluxDB](../managing-workflow/platform-monitoring.md#configuring-off-cluster-influxdb) and [Grafana](../managing-workflow/platform-monitoring.md#off-cluster-grafana)
 
+#### (Experimental) Kubernetes Native Ingress
+
+Workflow now offers [experimental native ingress](experimental-native-ingress.md) to take advantage of native Kubernetes routing. Any compatible Kubernetes ingress controller can be used in place of Workflow's nginx-based deis-router. Follow [this guide](experimental-native-ingress.md) to enable experimental native ingress.
 ## Add the Deis Chart Repository
```
bacongobbler: Can you add an extra newline here between the paragraph and the new heading? Just a small pet peeve of mine :)

charts/workflow/requirements.yaml
krancour: So the plan is to include the router even if the user opts in to native ingress?

src/installing-workflow/experimental-native-ingress.md
krancour: Does "separate and clear" become a maintenance burden if / as additional flags for major / experimental features get added. I'm afraid of committing to the notion of a page per permutation.

charts/workflow/requirements.yaml
mboersma: No, we wouldn't include deis-router in that case. The debate is whether to use Helm 2.2.x's tags & conditions feature as the mechanism to disable deis-router, or whether to use our convention of big `{{ if }}` blocks which is more backward-compatible and consistent with other components.

I want to argue that we turn off deis-router "the old way," then make a separate effort to refactor all cases to use the new Helm functionality. But it may not be that simple; see https://github.com/deis/workflow/issues/771.

src/installing-workflow/experimental-native-ingress.md
kris-nova: I think the wise thing to do here is to let problems present themselves instead of trying to be too defensive and predict the future. If it becomes a problem, we can always change. The concern seems trivial to be honest, and this an `experimental` feature after all, shouldn't we be `experimenting` to see what works best?

kris-nova:
@bacongobbler thanks for the review, your concerns have been addressed.

bacongobbler:
As far as I'm concerned this all looks goof to me.

charts/workflow/requirements.yaml
krancour: @mboersma I don't like the "big {{ if }}" approach because it requires changes to the router chart that endow that chart with an awareness that some parent chart might include it but wish for it to do nothing. That doesn't seem very clean to me.

src/installing-workflow/experimental-native-ingress.md
krancour: I don't think it's a trivial concern. We generally accept the practice of avoiding DRY exceptions in code. imo, docs shouldn't be any different. In either case, a DRY exception represents an on-going maintenance burden.

Though I don't think it's a trivial concern, I'm not going to belabor it either. You're doing the legwork on this, so I defer to your judgement.

charts/workflow/requirements.yaml
mboersma: It's not very clean, agreed. The advantages are:
- works with older versions of `helm`
- matches the (ugly) way we do things in other charts

But I was off-base in suggesting we could fix everything in a unified way: it may not be feasible to convert other sub-charts to use tags & conditions in a perfectly backward-compatible fashion. So this should be an isolated decision. I know you feel strongly about it @krancour and my main concern at this point is that @kris-nova doesn't get whiplash if we want to revert the deis-router changes. 😄

kris-nova:
Wooo! Thanks @mboersma and @bacongobbler for the approval here!

I think we should wait for the other PRs to fall into alignment before merging. Particularly https://github.com/deis/controller/pull/1243

mboersma:
Jenkins, add to whitelist.

mboersma:
Replaced by #1274.

PR #745, opened 2017-02-25
docs(managing-workflow) Document k8s annotation for AWS ELB idle timeout

felixbuenemann:
This updates the docs to describe how to persist the AWS ELB idle timeout by using the proper k8s service annotation instead of following the manual instructions, which get reset if k8s re-configures the ELB.

The annotation was added in kubernetes/kubernetes#30695 and merged targeting k8s v1.4 in August 2016. I have verified that it works as expected on k8s v1.4.6.

### DIFF
--- a/src/managing-workflow/configuring-load-balancers.md
+++ b/src/managing-workflow/configuring-load-balancers.md
@@ -6,7 +6,15 @@ Depending on what distribution of Kubernetes you use and where you host it, inst
 
 If a load balancer such as the one described above does exist (whether created automatically or manually) and if you intend on handling any long-running requests, the load balancer (or similar) may require some manual configuration to increase the idle connection timeout.  Typically, this is most applicable to AWS and Elastic Load Balancers, but may apply in other cases as well.  It does not apply to Google Container Engine, as the idle connection timeout cannot be configured there, but also works as-is.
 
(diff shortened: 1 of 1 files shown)


src/managing-workflow/configuring-load-balancers.md
```diff
@@ -6,7 +6,15 @@ Depending on what distribution of Kubernetes you use and where you host it, inst
 
 If a load balancer such as the one described above does exist (whether created automatically or manually) and if you intend on handling any long-running requests, the load balancer (or similar) may require some manual configuration to increase the idle connection timeout.  Typically, this is most applicable to AWS and Elastic Load Balancers, but may apply in other cases as well.  It does not apply to Google Container Engine, as the idle connection timeout cannot be configured there, but also works as-is.
 
-If, for instance, Deis Workflow were installed on kube-aws, this timeout should be increased to a recommended value of 1200 seconds.  This will ensure the load balancer does not hang up on the client during long-running operations like an application deployment.  Directions for this can be found [here](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/config-idle-timeout.html).
+If, for instance, Deis Workflow were installed on kube-aws, this timeout should be increased to a recommended value of 1200 seconds.  This will ensure the load balancer does not hang up on the client during long-running operations like an application deployment.
```
kris-nova: Is 1200 seconds the longest AWS allows? I think we should suggest the maximum value here, and explicitly call out that we are doing so. Just my thoughts from going through this in kops.

felixbuenemann: No, it is not the longest, but the deis-controller times out request after 20 minutes, which is why the docs recommend to set the same timeout. According to the AWS docs the value must be between 1 and 3600 seconds.

PR #746, opened 2017-02-28
feat(configuring-load-balancers): use kubernetes idle timeout annotation

bacongobbler:
This annotation is undocumented, but was introduced in https://github.com/kubernetes/kubernetes/pull/30695. It has been available since k8s v1.4.


bacongobbler:
closing in favour of #745

Issue #667, opened 2016-12-19
Remove the workflow-upgrade job


kmala:
The upgrade job(https://github.com/deis/workflow/blob/master/charts/workflow/templates/upgrade-job.yaml) needs to removed from the charts once the helm feature https://github.com/kubernetes/helm/pull/1703 is released and docs need to updated regarding the same.

mboersma:
This can be closed when Helm v2.2.0 is released.

PR #748, opened 2017-03-01
ref(chart): remove upgrade-job.yaml

vdice:
Closes https://github.com/deis/workflow/issues/667


vdice:
Quick smoke test of upgrade from latest Workflow chart version (v2.11) to this pr chart version (workflow-v2.11.1-20170301205052-sha.d6d39f5) using off-cluster object storage was successful with a few example apps.

PR #749, opened 2017-03-02
docs(tuning-component-settings): elaborate with practical examples

vdice:
cc @nathansamson


vdice:
Good suggestion @bacongobbler -- updated.

PR #756, opened 2017-03-07
ref(charts): disable router/host_port/enabled

bacongobbler:
On some cloud providers like GKE, it acts very wonky with host_ports enabled by default.

cc @slack


bacongobbler:
closing in favour of #757 since it's more in-depth.

PR #757, opened 2017-03-07
ref(router): disable router hostPort usage by default

slack:


bacongobbler:
ping @tmc, this is what was causing the GKE service loadbalancer wonkiness with your cluster. Disabling host_port for your router service should fix it for you.

tmc:
Danke.

PR #770, opened 2017-03-10
docs(builder-nodeselector): add BUILDER_POD_NODE_SELECTOR document

zinuzoid:
refs https://github.com/deis/builder/pull/492

Here's how it looks.
screenshot 2017-03-10 15 04 37

### DIFF
--- a/src/managing-workflow/tuning-component-settings.md
+++ b/src/managing-workflow/tuning-component-settings.md
@@ -50,9 +50,10 @@ limits_memory = "50Mi"
 
 The following environment variables are tunable for the [Builder][] component:
 
-Setting | Description
-------- | ---------------------------------
-DEBUG   | Enable debug log output (default: false)
+Setting                     | Description
+--------------------------- | ---------------------------------
+DEBUG                       | Enable debug log output (default: false)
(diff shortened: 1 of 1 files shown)


src/managing-workflow/tuning-component-settings.md
```diff
@@ -50,9 +50,10 @@ limits_memory = "50Mi"
 
 The following environment variables are tunable for the [Builder][] component:
 
-Setting | Description
-------- | ---------------------------------
-DEBUG   | Enable debug log output (default: false)
+Setting                     | Description
+--------------------------- | ---------------------------------
+DEBUG                       | Enable debug log output (default: false)
+BUILDER_POD_NODE_SELECTOR   | A node selector setting for builder job. As builder job sometimes consume a lot of node resources. In case you want a builder job to run in specific node only, so it won't effect the critical node. for example `pool:testing,disk:magnetic`
```
vdice: Minor grammatical suggestion -- perhaps change to something like:

>As it may sometimes consume a lot of node resources, one may want a given builder job to run in a specific node only, so it won't affect critical nodes.

Otherwise, thank you for adding!

PR #782, opened 2017-03-30
feat(ingress): experimental native ingress

mboersma:
This is just a rebase of #732 to see if we can avoid the CI issues it was running into. All credit goes to @kris-nova of course.

Work for experimental native ingress. Run the documentation to see how to use the new feature.

Requires deis/controller#1243
Requires deis/builder#495
Requires deis/router#316


mboersma:
Since this is just a rebasing of #732, I'm going to consider those LGTMs also to apply here. 😅

Issue #741, opened 2017-02-15
`deis addons` command should work with k8s service brokers


mboersma:
Developers should be able to list, provision and deprovision services available to them through the deis workflow cli. This is forward-looking, anticipating a service broker implementation that eventually comes out of k8s sig-service-catalog efforts, or [steward](https://github.com/deis/steward).

This could be implemented as a deis CLI plugin, if we want to keep things decoupled or vendor-neutral.

Cryptophobia:
This issue was moved to teamhephy/workflow#34

PR #791, opened 2017-04-10
doc(applications): deploying add-ons

bacongobbler:
This is the design document for `deis addons` and is currently a work-in-progress. This document is expected to change over time as work is being done.

closes https://github.com/deis/workflow/issues/741 once the work has been laid out and implemented.


src/applications/deploying-addons.md
```diff
@@ -0,0 +1,102 @@
...
+```
+
+Depending on the plan, some services can take some time before it becomes available for use. Use
+`deis addons:wait` to wait for the service to become available.
+
+Once the service has been attached to the application, a DATABASE_URL setting will be available in
+the application's configuration and will contain the URL used to access the newly provisioned
+service.
+
+You can choose the alias that the add-on uses on the application using the `--as` flag. This will
```
ultimateboy: How would this work with a broker that has multiple credential values, such as `DB_USER` and `DB_PASS`?

bacongobbler: This is a very good question. This is a straight cut from Heroku's docs and I'm not sure if it's a resonable effort to do this for a first pass. I'm not sure how Heroku handles this either as I know they have add-ons like Papertrail with multiple envvars. I'll go play around with some of their add-ons and see what works and what doesn't.

bacongobbler:
added `deis addons:services` to the docs to list the service catalog.

src/applications/deploying-addons.md
bacongobbler: I was thinking we could do some crazy docker-like mapping in the form of `--as DATABASE_URL=DB_URL,DATABASE_USER=DB_USER`, but again I'm concerned about making the translation overly complex. I'm sure this will come into play as things start being implemented.

bacongobbler:
removed the `--as` support in the doc. Heroku's implementation is shoddy at best, and I couldn't think of a perfect solution out the door, so for an initial implementation we'll just accept whatever envvars the service broker gives us.

krancour:
@bacongobbler, I think one needs to carefully read docs or interactive help to understand that `addons:list` is provisioned and bound services while `addons:services` are _available_ services. Would the intent of each of those list operations be less ambiguous if we used `addons:provisioned` (or `addons:installed`) and `addons:available`, respectively?

bacongobbler:
I was trying to make it a 1:1 mapping with `heroku` CLI commands, but I'd be amenable to `deis addons:list` being the catalog list and `deis addons:installed` being a list of addons provisioned under the user's account. How about that?

krancour:
Is parity with Heroku still a concrete objective? In either event, I'm happy with what you have just proposed.

mboersma:
> Is parity with Heroku still a concrete objective?

Yes, because it's still a familiar toolset to many people, and that's still the general inspiration for Deis Workflow functionality. But it's not set in stone, if there's a good argument for doing things a different way--and it sounds like there might be here.

bacongobbler:
I think this is a very good opportunity to change the heroku CLI names for specific commands. I agree that `addons:list` and `addons:services` is too confusing. I'll update the docs with the commands I proposed.

bacongobbler:
So as it turns out, instances are namespaced. Makes sense, but we haven't had to provision user-namespaced resources before... Everything was either an annotation on the application's service or on the router.

What I think that means going forward is that we can introduce a `myuser` namespace creation at `deis register` (along with a data migration for existing users), then each instance that is provisioned from a broker is created in that user's namespace. Instance bindings just reference instances by name which is globally unique, so having the instance in one namespace and the binding in another shouldn't matter, right? This would also have the added benefit that all instances are removed when the user is deleted. This would also be a useful transition into `deis teams`, in that instances are deployed to a team's namespace.

Any opinions/thoughts on the matter, @krancour or @ultimateboy?

bacongobbler:
From what it looks like, my assumption was wrong. instances and instance bindings must exist in the same namespace, so instances must be bound to the application then. That kinda sucks. I'll file a ticket upstream and see if there's a way we can work around this.

As seen here, I have an instance in the `bacongobbler` namespace...

```
><> kubectl --context=service-catalog get instances --namespace bacongobbler
NAME           KIND
ups-instance   Instance.v1alpha1.servicecatalog.k8s.io
```

...And an instance binding in the `test-ns` namespace following the basic walkthrough.

```
><> kubectl --context=service-catalog get bindings -n test-ns ups-binding -o yaml
apiVersion: servicecatalog.k8s.io/v1alpha1
kind: Binding
metadata:
  creationTimestamp: 2017-04-12T22:44:01Z
  finalizers:
  - kubernetes
  name: ups-binding
  namespace: test-ns
  resourceVersion: "76"
  selfLink: /apis/servicecatalog.k8s.io/v1alpha1/namespaces/test-ns/bindings/ups-binding
  uid: 85e8d1d6-1fd1-11e7-b4a0-0242ac110006
spec:
  instanceRef:
    name: ups-instance
  osbGuid: 17633cb4-dff1-4221-909d-e3dadc2f3707
  secretName: my-secret
status:
  conditions:
  - message: The binding references an Instance that does not exist. Binding "test-ns/ups-binding"
      references a non-existent Instance "test-ns/ups-instance"
    reason: ReferencesNonexistentInstance
    status: "False"
    type: Ready
```

bacongobbler:
You know what, the idea of the instance and the instance binding living in the same namespace makes more sense anyways because in a team-centric world, all apps will be deployed in the same namespace as the team name so the instance should live in there.

I'll be white-boarding this more overnight and try to come up with a better solution tomorrow.

src/applications/deploying-addons.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Detaching the Add-on
+
+Similarly, an add-on can be unbound from an application, but still be available to the user for
+future binding to another application.
+
+```
+$ deis addons:detach deis-postgresql:standard --app wooden-rowboat
+Detaching deis-postgresql:standard from wooden-rowboat... done, v6
+```
```
arschles: what do you want to happen when a detach (unbind) happens before a deprovision?

src/applications/deploying-addons.md
```diff
@@ -0,0 +1,101 @@
...
+A user can use `deis addons:installed` to see if they have a database provisioned and what plan it
+is using.
+
+```
+$ deis addons:installed | grep -i postgresql
+$ deis-postgresql:standard
+```
+
+If the application doesn’t yet have a database provisioned, a user can create a new database using
+the CLI.
```
arschles: This is a little confusing. Does this phrase:

```
a user can create a new database using the CLI
```

Refer to a provision operation?

src/applications/deploying-addons.md
```diff
@@ -0,0 +1,101 @@
...
+allows applications to consume services like databases that exist somewhere via common environment
+variables like `DATABASE_URL`.
+
+## Listing Available Add-ons
+
+A user can use `deis addons:installed` to see if they have a database provisioned and what plan it
+is using.
+
+```
+$ deis addons:installed | grep -i postgresql
```
arschles: How does someone get a new addon into this list? In service-catalog land, they'd do that by creating a `Broker` resource. Is there an analog with the `deis` CLI?

src/applications/deploying-addons.md
```diff
@@ -0,0 +1,101 @@
+# Deploying Add-ons
+
+The goal of `deis addons` is to give users the power to provision consumable services such as
+a postgres database, a minio bucket, or a logstash connection) to their applications. `deis addons`
```
arschles: The closing `)` doesn't have an opening one to match.

src/applications/deploying-addons.md
```diff
@@ -0,0 +1,101 @@
...
+with [service brokers][] to the Kubernetes ecosystem via the [Open Service Broker API][].
+
+Users of Workflow use `deis addons` to provision an add-on offered by [service brokers][] for
+their applications. The end-goal is to provide a way for users to consume services from brokers and
+have their applications use those services without needing detailed knowledge about how those
+services are created or managed.
+
+As an example, most applications deployed by Workflow need a data store of some kind. `deis addons`
+allows applications to consume services like databases that exist somewhere via common environment
+variables like `DATABASE_URL`.
```
arschles: Is there a way for the app developer (the person who executes `deis addons:attach` for example) to tell what env vars are available, after they do the attach operation (note there's no way in service-catalog land to tell what's available before the attach)?

bacongobbler: I know that Heroku exposes them in `config:list`. I'm sure we can do the same somehow.

src/applications/deploying-addons.md
bacongobbler: There is not. The expectation is that we will be listing the service classes, but the brokers themselves is managed by service-catalog and we are not adding any administration to service catalog. Just communicating with it similar to how we consume the kubernetes API.

bacongobbler: So basically if an admin creates a new broker resource, it will populate new service classes into the mix for users to query via `deis addons:catalog`

I haven't added addons:catalog into the docs yet but I am going to amend these once again so users can list instances, bindings and service classes.

src/applications/deploying-addons.md
bacongobbler: Correct. It relates to an instance resource being created on the service-catalog's side.

Basically there are three operations:

 - list the catalog
- CRUD the instances
- CRUD the instance bindings

src/applications/deploying-addons.md
bacongobbler: So let's think about this in a team-oriented fashion:

 - I have a team (a namespace to store all apps deployed by a team)
 - I have many apps deployed in this namespace
 - I have a few instances running in this namespace

I would expect that, as a user, I can bind an app to the instance as well as unbind apps from the instance.

This is essentially useless for the "instance in the app's namespace" use case though. Again I'll give this some more thought. :)

krancour:
@bacongobbler whether it should be possible to bind to instances in other namespaces was the subject of lots and lots of discussion very early on, but not so much as of late. I am of the opinion that in large enterprises and in microservice architectures, the notion of n applications all binding to a common instance that _belongs_ to none of them (but is managed as its own thing) is probably not an uncommon use case. But that's not something that service catalog currently does because it would simply defy k8s existing permissions model and I don't think we were about to go re-inventing that.

The practical implication of that is exactly what you seem to have concluded already-- that service instances and bindings for those instances should all be provisioned within an app's own namespace (or within a team namespace that contains multiple apps? is that a thing now?)

Also, since Workflow is very app centric, would provision + bind be combined into a single step for Workflow users?

src/applications/deploying-addons.md
```diff
@@ -0,0 +1,101 @@
...
+
+If a user wishes to see all available addons, they can use `deis addons:list`.
+
+```
+$ deis addons:list
+slug             name
+---------------  ----------------------
+bonsai           Bonsai Elasticsearch
+deis-postgresql  Deis Workflow Postgres
+librato          Librato
```
bacongobbler: Yep! The `slug` and `name` field relate directly with a service class' `brokerName` and `description` fields. I'll change the `name` field to `description` and `slug` to `name` to hopefully make things clearer. This is just how Heroku calls things.

bacongobbler:
> or within a team namespace that contains multiple apps? is that a thing now?

Nope! Just a twinkle in my eye.

> Also, since Workflow is very app centric, would provision + bind be combined into a single step for Workflow users?

Yes. Honestly I've just revamped the documentation again to only have 5 commands:

 - addons:list (list instances provisioned for this app)
 - addons:create (create an instance + binding, using format `serviceclassname:plan` as argument)
 - addons:destroy (destroy an instance + binding using service class name as the argument)
 - addons:catalog (list available service classes)
 - addons:plans (list available plans for the given service class name)

`addons:plans` is still yet to be added to this PR, but will be up in a sec.

src/applications/deploying-addons.md
bacongobbler: So given the `ups` broker is installed, `deis addons:catalog` would look like:

```
name                             description
---------------------------   -----------------------------
user-provided-service   User Provided Service
```

src/applications/deploying-addons.md
```diff
@@ -0,0 +1,74 @@
...
+
+For example, to provision a `standard` plan database:
+
+```
+$ deis addons:create deis-postgresql:standard --app wooden-rowboat
+Creating deis-postgresql:standard... done
+Attaching deis-postgresql:standard to wooden-rowboat... done, v5
+```
+
+Once the instance has been attached to the application, a DATABASE_URL setting will be available in
```
krancour: "setting" is a bi ambiguous. Change to "environment variable"?

krancour:
@bacongobbler I like where this is headed. It's clean and simple. There's a lot more complexity that _could_ be exposed, but given Workflow's Heroku-esque focus on 12factor apps, this looks to be just right. 💯

src/applications/deploying-addons.md
bacongobbler: changed.

bacongobbler:
k, `deis addons:plans` has now been added as well.

src/applications/deploying-addons.md
bacongobbler: added more context lower down; should be available through `deis config:list`.

PR #801, opened 2017-04-25
Enable registry-proxy for CNI users

Bregor:
PR to allow usage of https://github.com/deis/registry-proxy/pull/11

### DIFF
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -57,6 +57,16 @@ global:
   # - true: deis-router will not be deployed. Workflow will not be usable until a Kubernetes ingress controller is installed.
   # - false: deis-router will be deployed (default).
   experimental_native_ingress: false
+  # If the Kubernetes cluster uses CNI
+  # use_cni: true
+  # Set the `listen` variable for registry-proxy's NGINX
+  #
+  # Valid values are:
+  # - 80: If the Kubernetes cluster run on GKE or AWS, or uses flannel or kubenet as a pod network
+  # - 127.0.0.1:5555: If the Kubernetes cluster uses CNI
+  #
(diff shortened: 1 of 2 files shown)


bacongobbler:
Would you mind adding additional documentation so others know how to use this feature? Thanks! :)

Bregor:
@bacongobbler you mean more comments in manifests or in some other place?

bacongobbler:
I'm meaning in deis.com/docs/workflow. If you could add something in [Production Deployments](https://github.com/deis/workflow/blob/master/src/managing-workflow/production-deployments.md) on running Workflow with a Kubernetes cluster with CNI and how to configure this feature in there, that'd be fantastic.

Bregor:
OK, sure. Will do it shortly.

Bregor:
@bacongobbler should I add example right to https://github.com/deis/workflow/blob/master/src/managing-workflow/production-deployments.md, or make a reference to https://github.com/deis/distribution/blob/master/docs/configuration.md and fork those repo too?

charts/workflow/values.yaml
```diff
@@ -57,6 +65,8 @@ global:
   # - true: deis-router will not be deployed. Workflow will not be usable until a Kubernetes ingress controller is installed.
   # - false: deis-router will be deployed (default).
   experimental_native_ingress: false
+  # If you're with CNI
```
vdice: maybe rephrase this comment as `# If the Kubernetes cluster uses CNI`

vdice:
@Bregor I think the first link (https://github.com/deis/workflow/blob/master/src/managing-workflow/production-deployments.md) which is here in the same repo, is fine for now.

Bregor:
Docs added in https://github.com/deis/workflow/pull/801/files#diff-e00bc5e4152371ec0f45a782e3235b87

PR #805, opened 2017-05-02
minor correction to the registration url

genecyber:


mboersma:
Could you also update the last line of that section to say:
```You are admin at http://deis.($hostname)```

mboersma:
Actually, I'm not sure this change is necessary. Immediately before this section the docs say:
> The `$hostname` value can be calculated by prepending `deis.` to the value set in `controller.platform_domain`.

I think that covers it, so I'm going to close this.

PR #819, opened 2017-06-01
docs(installing-workflow): helm init upgrade note

vdice:
`helm init --service-account=foo --upgrade` supported as of v2.4.2


src/installing-workflow/index.md
```diff
@@ -30,14 +30,7 @@ $ kubectl create clusterrolebinding helm --clusterrole=cluster-admin --serviceac
...
 
-If `helm` is already installed in cluster without sufficient rights, the only way for now is to reinstall it:
-
-```
-$ kubectl delete deployment tiller-deploy -n kube-system
-$ kubectl create sa tiller-deploy -n kube-system
-$ kubectl create clusterrolebinding helm --clusterrole=cluster-admin --serviceaccount=kube-system:tiller-deploy
-$ helm init --service-account=tiller-deploy
-```
+If `helm` is already installed in cluster without sufficient rights, simply add `--upgrade` to the last command above.
```
sgoings: instead of "to the last command above" maybe just say "to the `init` command above"?

PR #821, opened 2017-06-05
docs(*): updates/changelog for v2.15.0 release

mboersma:
See also #822--some docs were commented out until we have a fix for chart verification.

### DIFF
--- a/src/roadmap/releases.md
+++ b/src/roadmap/releases.md
@@ -133,6 +133,9 @@ If the component has a correlating [Kubernetes Helm][] chart,
 this chart will also be packaged, signed and uploaded to its production chart repo.  Please
 verify it can be fetched (and verified):
 
+!!! note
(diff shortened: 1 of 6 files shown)


src/roadmap/releases.md
```diff
@@ -133,11 +133,14 @@ If the component has a correlating [Kubernetes Helm][] chart,
 this chart will also be packaged, signed and uploaded to its production chart repo.  Please
 verify it can be fetched (and verified):
 
+!!! note
+    Verification of chart provenance is [currently broken](https://github.com/deis/workflow/issues/822).
```
kris-nova: `s/broken/unsupported/g`? Broken seems like a strong word. Thoughts?

mboersma: You're right--"broken" implies that we want to fix it ASAP, which I *think* is true. This particular document is public, but not really user-facing (it's our instructions to ourselves about how to do a Workflow release), so I think it's ok to say "broken" here.

PR #850, opened 2017-08-09
Add instructions for DNSmasq

### LINKED ISSUE #833: nip.io not filtered by my DNS provider
My DNS provider filters local IPs, so this prevented *.nip.io address to work correctly and caused the following error instead :

```
➜  ~ curl http://deis.192.168.64.2.nip.io/v2/ && echo
curl: (6) Could not resolve host: deis.192.168.64.2.nip.io
```

I first tried to use `xip.io` instead before realizing this sas caused by my DNS provider which, of course, failed to solve the problem.

I could thinks of these workarounds :

- In documentation, in case of failure with *.nip.io, request the user to change DNS provider and use another provider such as Google and update DNS settings to something like 8.8.8.8 / 8.8.4.4.
- Add entries to `/etc/hosts`

Maybe there are more friendly solutions. However, I am not sure these work-around are OK for all use-cases.

ybart:
Closes #833.

### DIFF
--- a/src/quickstart/provider/minikube/dns.md
+++ b/src/quickstart/provider/minikube/dns.md
@@ -15,9 +15,11 @@ $ minikube ip
 
 ## Prepare the Hostname
 
-Now that you have the ip address of your virtual machine, we can use the `nip.io` DNS service to
+Now that you have the ip address of your virtual machine, we can use the `nip.io` DNS service or `dnsmasq` to
 route arbitrary hostnames to the Deis Workflow edge router. This lets us point the Workflow CLI at
 your cluster without having to either use your own domain or update DNS!
+
+### Using `nip.io`
 
 To verify the Workflow API server and nip.io, construct your hostname by taking the ip address for
 your load balancer and adding `nip.io`. For our example above, the address would be `192.168.99.100`.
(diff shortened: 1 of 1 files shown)


mboersma:
Jenkins, add to whitelist.

src/quickstart/provider/minikube/dns.md
```diff
@@ -31,6 +33,34 @@ $ host something-random.192.168.99.100.nip.io
 something-random.192.168.99.100.nip.io has address 192.168.99.100
 ```
 
+### Using DNSMasq
+
+If you `nip.io` is working for you, you can skip this section, and proceed to verify the hostname.
```
vdice: A very minor grammatical typo here -- perhaps should read `If nip.io is working for you...` (can remove the extra `you`)

src/quickstart/provider/minikube/dns.md
```diff
@@ -31,6 +33,34 @@ $ host something-random.192.168.99.100.nip.io
...
+
+If you prefer not to use `nip.io` or cannot (because your DNS provider might have blocked it), you can use `dnsmasq` on Linux and macOS or `Acrylic` on Windows.
```
mboersma: `dnsmasq` didn't work out-of-the-box for me on macOS until I manually added 127.0.0.1 to my DNS servers list--I gather this is a common issue when getting DNS via DHCP. Maybe it's worth adding a sentence here like:
```
You may need to ensure that the `dnsmasq` service at 127.0.0.1 is listed as a DNS server for your network connection.
```

mboersma:
I'm going to go ahead and merge this and follow up with our ticky-tack changes in a separate PR. Thanks @ybart!

PR #855, opened 2017-08-22
docs(README.md): add notice of final release and maintenance mode

mboersma:
If this wording is appropriate, I'll make a set of cookie-cutter PRs to include this in all the relevant repositories.


mboersma:
Here's another version, based on feedback from @sgoings:

Issue #481, opened 2016-09-01
Configurable endpoint for S3 object storage


markkwasnick:
Deis currently supports Minio (S3-compatible API) for object storage only within the cluster and not for production deployments. Deis also supports S3 for object storage, but not with a configurable "regionendpoint", requiring the use of specific Amazon AWS regions.

Request is to modify Deis to accept a s3.region_endpoint parameter within the generate_params.toml file on the deis/charts project, which will be stored as a Kubernetes secret like the other parts of configuration. This will allow for operators to configure Deis Object Storage to use hosted S3-compatible APIs outside of the cluster.

(Related) Is a change to the deis/object-storage-cli to pass in the configurable endpoint above to the s3 driver via the optional "regionendpoint" param.

(per doc: https://github.com/docker/distribution/blob/master/docs/storage-drivers/s3.md)

markkwasnick:
Adding a note here: Probably also want to include an option to toggle secure yes/no for the docker storage driver.

bacongobbler:
If anyone has a good doc on getting this set up, that would be appreciated. We can fizzle that down into a consumable document for other users.

Cryptophobia:
This issue was moved to teamhephy/workflow#52

PR #857, opened 2017-08-29
WIP: declare "off_cluster_minio"

kingdonb:
for an External minio server using s3.accesskey and s3.secretkey,
will add s3.regionendpoint later

WIP: to resolve deis/workflow#481

More changes will go in deis/builder deis/controller deis/database
deis/registry, all of which have dependencies on s3/swift storage

In PoC I overrode DEIS_MINIO_SERVICE_HOST and DEIS_MINIO_SERVICE_PORT

I'd like to use that shortcut and it seems safe, but I'm open to doing it another way if it's cleaner.  I would decompose host and port from value in s3.regionendpoint, reorganizing values.yaml so the off-cluster minio configuration is very obvious.

I could not reliably use Deis Workflow on any of my (relatively not-top-of-the-line) portable machines via minikube or minishift without adding a remote storage server for minio, and adding some persistent disk.  Can we get the configuration for this mode worked out in 2.19 final release?

### DIFF
--- a/charts/workflow/values.yaml
+++ b/charts/workflow/values.yaml
@@ -9,6 +9,7 @@ global:
   # - gcs: Store persistent data in Google Cloud Storage
   # - minio: Store persistent data on in-cluster Minio server
   storage: minio
+  off_cluster_minio: true
 
   # Set the location of Workflow's PostgreSQL database
   #


kingdonb:
Values.global.storage "minio" <-- is this value read and used anywhere other than in `charts/minio`?

If not, we can just pick a different value instead of "minio" so that it will be distinct from internal minio.

https://gist.github.com/kingdonb/845945908336cbeefb3e8cce73ec5dc2

These changes were all I needed to enable off-cluster minio on my local/minikube Workflow instance.

kingdonb:
I thought I had better get some feedback before I went to the trouble and opened four separate pull requests on [deis/builder](/deis/builder), [deis/registry](/deis/registry), [deis/database](/deis/database), and [deis/controller](/deis/controller)... please take a look at my patch and suggest alternatives if you think it is not in good shape to merge.  The better approach is probably not to try overwriting kubernetes service host and port variables... but I am new at this.

kingdonb:
This will be better if I follow the guidance in #481 more closely, namely the part about setting `s3.regionendpoint`

mboersma:
Jenkins, add to whitelist.

kingdonb:
Related to deis/monitor#191

If you want to support a cloud-neutral storage for swift objects, to be complete would also need to provide an answer for block storage or at least for the storage requirements of Grafana and InfluxDB

kingdonb:
Jenkins, test this please

